PR #4, opened 2017-04-18
quotations: strings in {{ }} delegated to 3rd party

gares:
See  https://github.com/LPCIC/elpi/blob/quotations/LANGUAGE.md for an example,
or https://github.com/LPCIC/coq-elpi/blob/master/theories/test_quotation.v for a more
complex use case.


gares:
The code also cleans up a little the `Compiler` module (`stack_term_of_ast` and co).

PR #7, opened 2017-06-15
Fix compilation on NixOS.

Zimmi48:
FTR the bench still doesn't work but this doesn't matter.


gares:
Thanks.

The bench contains various tests that do fail (JFYI).  It really needs some care

PR #10, opened 2017-11-10
CHR: new syntax and implementation

gares:
From the user perspective:
- patterns operate on frozen terms
- two alignment mode
  - auto for L\lambda
  - manual:
    - matched constraints are spread
    - one chooses in which context the new goal is injected
    - injection is manual
- less bugs, better error messages

Inside the engine:
- the new code is more clear
  - but makes many passes on the terms to shift them correctly
- all UVar that intervene are expanded to level 0
  - this may resume other goals that then would immediately
    be suspended

### DIFF
--- a/ELPI.md
+++ b/ELPI.md
@@ -295,15 +295,17 @@ mode (foo i o).
 
 ## Syntactic constraints
 
-A goal can be suspended on a list of variables with the `declare_constraint` built in.
+A goal can be suspended on a list of variables with the `declare_constraint` 
+built in.  In the following example the goal `even X` is suspended on the
+variable `X`.
 ```prolog
 goal> declare_constraint (even X) [X].
 Success:
 Constraints:
-   ⊢ (even X)
+  even X  /* suspended on X */
 ```
-Suspended goals are resumed as soon as any of variables they are suspended on
-gets assigned.
+Suspended goals are resumed as soon as any of the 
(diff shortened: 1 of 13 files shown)


gares:
@sacerdot please give a look at ELPI.md to see if the changes I propose make sense.

In particular the old code was leaving the new goal in the disjoint union of the contexts, that is probably sound but not what one wants.  I believe it is better to do a substitution at the meta level (using `copy` since one replaces constants for terms) than asking the main program to do it, e.g. via a bunch of spurious letins (as we were doing). This is the main visible change of this patch.

Another change is that I've introduced the notion of main key (the first variable on which a goal is suspended).  Only goals suspended on the main key are matched together.  It is OK for the use cases we have in mind.  I'm not very happy with it, but it the best I could get.
The benefit is that such check (same key) is now done very early (it was done in the alignment phase) speeding up things (and especially cleaning up the traces from spurious attempts).  Also, I pre-compute the list of head symbols of each rule, so that I can further trim down the set of candidates. Finally I now freeze just before the matching, so that one only writes terms at the meta level: `uvar K L` and no more `?? as X` with the limitation that `?? K L` is not allowed since we don't know yet `L` because it is the freezing operation that gives it... and one was forced to do `(X = uvar K L)` in the guard... It looks more clear now, IMHO.

PR #24, opened 2019-01-03
build using dune

gares:
This PR tries to build elpi with dune.

Current annoyances:
- ~https://github.com/ocaml/dune/issues/753 to remove `elpi_` from file name prefixes~ (happens to work after elpi 1.3)
- https://github.com/ocaml/dune/issues/61 to disable warn 55 on elpi_runtime_trace_on

Hackish workaround:
- https://github.com/ocaml/dune/issues/1212 to hack .merlin in order to load the desired ppx


src/dune
```diff
@@ -0,0 +1,48 @@
...
+    ((pps ppx_deriving.std) elpi_util elpi_ast elpi_data elpi_compiler)))
+  (libraries re.str camlp5.gramlib unix)
+  (flags -linkall)
+  (modules elpi elpi_util elpi_parser elpi_ast elpi_compiler elpi_data elpi_ptmap elpi_builtin elpi_API elpi_runtime_trace_on elpi_runtime_trace_off)
+)
+
+(rule
+  (targets elpi_runtime_trace_on.ml)
+  (deps elpi_runtime.ml)
+  (action (copy# %{deps} %{targets}))
```
rgrinberg: Slightly shorter is to have:

```
(rule (copy# elpi_runtime.ml elpi_runtime_trace_on.ml))
```

Since dune is aware of how `copy#` works, it's able to infer what the dependencies and targets are already.

src/dune
```diff
@@ -0,0 +1,48 @@
+(library
+  (name elpi)
+  (public_name elpi)
+  (preprocess (per_module
+    ((pps ppx_deriving.std trace_ppx --trace_ppx-on) elpi_runtime_trace_on)
+    ((pps ppx_deriving.std trace_ppx --trace_ppx-off) elpi_runtime_trace_off)
+    ((action (system "camlp5o -I . -I +camlp5 pa_extend.cmo pa_lexer.cmo %{input-file}")) elpi_parser)
```
rgrinberg: A bit curious about the necessity of `system` here. You don't seem to be using shell features.

elpi.install
```diff
@@ -0,0 +1,97 @@
+lib: [
```
rgrinberg: Checking the file in is wrong because it contains some platform specific stuff (`.cmxs` aren't necessarily available everywhere), but luckily it's harmless as dune will overwrite it anyway when building the package. I suggest you just remove this file and add it to your .gitignore.

gares:
Thanks

src/dune
gares: thanks

elpi.install
gares: I tried with `run`, but it insists in running `camlp5o.opt` that is then unable to load the `cmo`s.

src/dune
gares: I tried with `run`, but it insists in running `camlp5o.opt` that is then unable to load the `cmo`s.

PR #26, opened 2019-03-28
elpi 1.2

gares:
This PR includes all the changes I've in my pocket and that I'd like to release as elpi 1.2

The main user visible change is that an `std.` namespace is not part of elpi and includes a bunch of useful predicates. A few of them are injected in the main namespace: `if`, `fst` and `snd` as well as the `ff` and `tt` terms of type `bool` and the `pr` constructor of `pair`. The type checker is now emitting more warning for predicates without a type, unless they end in `.aux` (See for example `rev` and `rev.aux`). Breakage should be very minimal. The compiler is not much better at reporting errors (locations).
One can access names in a name space with the new directive `shorten foo.bar.baz` that lets you use `baz` in place of the full name. See ELPI.md for more info.

The main API change is that the functions to convert data in the FFI are now stateful (in the sense that they can also access the constraints store and company) so more datatypes can be converted using the FFI (instead of converting them as `any`, eg no conversion, and then doing things by hand). A few minor refinements to other APIs (especially related to locations) are there.

The `elpi` utility has a more structured `--help`.

The branch is a bit messy, since it also includes an extension to spilling that was a long waited for in coq-elpi.


gares:
I've successfully tested this in coq-elpi, bet before merging I'd like @sacerdot (and his collaborators) to test it.

PR #36, opened 2019-08-19
Restriction bug

gares:
This is mostly to make @sacerdot aware.

The commit history should be clear and I've added a test for each piece of code I've touched.

Note: the code containing "(*CSC: XXX why vardepth and not to_ ? *)" has been rewritten, and indeed we use `to_` all the time


gares:
Bottom line, I should really verify the `HO` module with Coq, it is too complex to be correct.

PR #52, opened 2020-01-17
[wip] Separate compilation

gares:
API (and some code) to let one do once and forall some compilation phases.
In particular the one unfolding quotation may use the host application state, so it should be done once and forall when the user wants it (and the result be stored and assembled later on)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,6 +18,7 @@ BUILD=_build/default
 SHELL:=/bin/bash
 TIMEOUT=90.0
 RUNNERS=\
+  dune \
   $(shell pwd)/$(INSTALL)/bin/elpi \
   $(addprefix $(shell pwd)/,$(wildcard _build/git/*/$(INSTALL)/bin/elpi.git.*)) \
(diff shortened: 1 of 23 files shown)


gares:
It is now possible, in the ML API, to parse, compile, and finally assemble programs.
Before one could only parse and compile+assemble.
Storing parsed files containing quotations is problematic, because host applications are in charge of unfolding them, and this may depend on their state. Now, by compiling, one forces this elaboration step to happen when he wants, and store compiled units (to be assembled and executed later).

FYI @sacerdot units cannot be compiled independently if one uses `local`. For now `local` is only accepted if one uses a single compilation unit (that can be composed of many files parsed together).
I'm not so sure if it is worth to fully implement `local` an separate compilation (one could "relocate" an already compiled unit by shifting it when all units are linked, and when the total number of locals is known).
I'm not implementing this now. The command line still uses the old API (all files are parsed and compiled in a single unit).

gares:
superseeded by #54

PR #53, opened 2020-01-24
Avoid build failures when the camlp5 binary changes

kit-ty-kate:
When elpi is built in one opam switch then rebuilt in another, the build fails because dune does not know camlp5 has changed.

Test case:
```
$ opam switch 4.08
$ dune build
$ opam switch 4.10
$ dune build
-     ocamldep src/.elpi.objs/parser.pp.mli.d (exit 2)
- (cd _build/default && /home/opam/.opam/4.10.0+trunk/bin/ocamldep.opt -modules -intf src/parser.pp.mli) > _build/default/src/.elpi.objs/parser.pp.mli.d
- >> Fatal error: OCaml and preprocessor have incompatible versions
- Fatal error: exception Misc.Fatal_error
-     ocamldep src/.elpi.objs/parser.pp.ml.d (exit 2)
- (cd _build/default && /home/opam/.opam/4.10.0+trunk/bin/ocamldep.opt -modules -impl src/parser.pp.ml) > _build/default/src/.elpi.objs/parser.pp.ml.d
- >> Fatal error: OCaml and preprocessor have incompatible versions
- Fatal error: exception Misc.Fatal_error
-       ocamlc src/.elpi.objs/byte/elpi__Builtin.{cmo,cmt}
```
This is because `(system "...")` is opaque from dune's point of vue and it cannot deduce the dependencies of the rule used to build `parser.ml`, which in this case is the `camlp5o` binary. To fix this we use the `(run ...)` command which is analysed by dune to deduce its dependencies.
Related issue: https://github.com/ocaml/dune/issues/3050

However, this PR does not work in dune currently as dune automatically appends the `.opt` suffix when available, this should be fixed in dune 2.2.0 hopefully, once https://github.com/ocaml/dune/pull/3051 has been merged.

### DIFF
--- a/src/dune
+++ b/src/dune
@@ -5,7 +5,7 @@
     ((pps ppx_deriving.std) API util ast data compiler)
     ((pps ppx_deriving.std elpi.trace_ppx --trace_ppx-on) runtime_trace_on)
     ((pps ppx_deriving.std elpi.trace_ppx --trace_ppx-off) runtime_trace_off)
-    ((action (system "camlp5o -I . -I +camlp5 pa_extend.cmo pa_lexer.cmo %{input-file}")) parser)
+    ((action (run camlp5o -I . -I +camlp5 pa_extend.cmo pa_lexer.cmo %{input-file})) parser)
     ))
   (libraries re.str camlp5.gramlib unix)
   (flags -linkall)
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,7 @@ install:
   - rm -rf $HOME/.opam
   - echo "installing cache"
   - cp -r $HOME/opam-cache/.opam $HOME/.opam
+  - opam update
 
 matrix:
   include:
--- a/dune-project
+++ b/dune-project
@@ -1,2 +1,2 @@
-(lang dune 1.6)
+(lang dune 2.2)
 (name elpi)
--- a/elpi.opam
+++ b/elpi.opam
@@ -22,7 +22,7 @@ depends: [
   "re" {>= "1.7.2"}
   "ANSITerminal" {with-test}
   "cmdliner" {with-test}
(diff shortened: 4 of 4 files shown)


kit-ty-kate:
I can't make sense of the CI failures. Could you look into that? I've tried several times but new errors keep coming

gares:
elpi and tests build fine. what fails is toyml (to become a proper test soonish), it seems to fail because of ppx_import
```
File "tests/sources/toyml/ast.ml", line 1, characters 11-37:

Error: show cannot be derived for [%import :Lexing.position]
```
Similarly the other jobs fail because ppx_import < ocaml 4.8.

I can get rid of that dependency easily, Lexing.position is not huge.

gares:
I've disabled the ocmpilation of toyml in master (and removed the dependency on ppx_import)

gares:
please rebase

kit-ty-kate:
Rebased and CI is now green.

gares:
Thanks

src/dune
```diff
@@ -5,7 +5,7 @@
     ((pps ppx_deriving.std) API util ast data compiler)
     ((pps ppx_deriving.std elpi.trace_ppx --trace_ppx-on) runtime_trace_on)
     ((pps ppx_deriving.std elpi.trace_ppx --trace_ppx-off) runtime_trace_off)
-    ((action (system "camlp5o -I . -I +camlp5 pa_extend.cmo pa_lexer.cmo %{input-file}")) parser)
+    ((action (run camlp5o -I . -I +camlp5 pa_extend.cmo pa_lexer.cmo %{input-file})) parser)
```
rgrinberg, line 8: Note that this is still a partial dependency specification. You aren't depending on the `cmo` files, and whatever else `camlp5` may be reading.

PR #54, opened 2020-01-28
Separate compilation

gares:
This is a massive refactoring making compilation purely functional and hence making the compilation unit not depend on the state of the system.

It also includes minor improvements all over the place, especially in the test suite


gares:
This time is for real CC @sacerdot , the code is satisfactory IMO, it only lacks a few comments/CHANGELOG.
I'll validate this branch on coq-elpi and hopefully merge and release 1.10.

Performance is not affected, although I had to add a quirk to allocate symbols in the *very same order* as before, otherwise the heuristics of grundlagen for constant unfolding fail (the use the constants internal number as a timestamp of their introduction in the typechecked environment... CC @gdufrc).
I'm fine with this for now, but it would be nicer if the timestamp was reified (elpi gives you a non-logical builtin that spits increasing integers... one could use it to implement the timestamp in a not less hackish way). FTR, with symbols allocated in the opposite order (the code was simpler to write that way) grundlagen is 3 times slower, so the heuristic is very useful ;-)

There is no command line option access to the separate compilation API, my use case is when you drive elpi from Coq and you store along with a .vo file a piece of elpi code. I used to store the AST, but that is not independent of Coq's state, since it may contain quotations (that in turn include Coq syntax). Compilation units do not depend on the host application state, since quotations are already desugared.

PR #77, opened 2020-08-03
fixes for camlp5 8.00

chetmurthy:
If you'd prefer, I can remove the changes #1, #2.  I wasn't sure I should put 'em in, but figured "better to put 'em in now and take 'em out later, it makes the committer's job easier"

Four sorts of fixes:

(1) Pervasives -> Stdlib (to allow clean compile with ocaml 4.10.0)

Maybe these should be removed?

(2) increase test timeout in Makefile (so tests can all pass)

Maybe remove?

(3) the type of Grammar rule has changed (there is now a "digest"
string to allow comparing actions for equality -- this was needed
because of bugs in left-factorization with such equality)

(4) the interface to Plexing has changed (instead of a function for
returning locations, there's a Location.t type -- needed to support
doc-comment extraction)

Both #3, #4 are superficial.

# Please enter the commit message for your changes. Lines starting
# with '#' will be ignored, and an empty message aborts the commit.
#
# On branch camlp5-8.00-patches
# Changes to be committed:
#	modified:   Makefile
#	modified:   elpi.opam
#	modified:   src/.ppcache/API.mli
#	modified:   src/.ppcache/ast.ml
#	modified:   src/.ppcache/compiler.ml
#	modified:   src/.ppcache/data.ml
#	modified:   src/.ppcache/runtime_trace_off.ml
#	modified:   src/.ppcache/runtime_trace_on.ml
#	modified:   src/ast.ml
#	modified:   src/builtin.ml
#	modified:   src/parser.ml
#	modified:   src/ptmap.ml
#	modified:   src/runtime.ml
#	modified:   src/util.ml
#

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ help:
 INSTALL=_build/install/default
 BUILD=_build/default
 SHELL:=/bin/bash
-TIMEOUT=90.0
+TIMEOUT=300.0
 RUNNERS=\
   dune \
   $(shell pwd)/$(INSTALL)/bin/elpi \
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ help:
 INSTALL=_build/install/default
 BUILD=_build/default
 SHELL:=/bin/bash
-TIMEOUT=300.0
+TIMEOUT=90.0
 RUNNERS=\
   dune \
   $(shell pwd)/$(INSTALL)/bin/elpi \
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ env:
     - OCAML_MIN=4.04.1
     - OCAML_MAX=4.09.0
     - PREDEPS="ocamlfind"
-    - DEPS="camlp5 ocamlfind ppx_deriving ppxlib re dune cmdliner ANSITerminal"
+    - DEPS="camlp5 ocamlfind ppx_deriving 'ppxlib<0.15.0' re dune cmdliner ANSITerminal"
     - MINDEPS="camlp5 ocamlfind dune re stdlib-shims"
     - JOBS=2
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ env:
     - OCAML_MIN=4.04.1
     - OCAML_MAX=4.09.0
     - PREDEPS="ocamlfind"
-    - DEPS="camlp5 ocamlfind ppx_deriving ppxlib<0.15.0 re dune cmdliner ANSITerminal"
+    - DEPS="camlp5 ocamlfind ppx_deriving ppxlib<0.15.0 re dune cmdliner ANSITerminal stdlib-shims"
     - MINDEPS="camlp5 ocamlfind dune re stdlib-shims"
     - JOBS=2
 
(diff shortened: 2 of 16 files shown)


Makefile
```diff
@@ -16,7 +16,7 @@ help:
 INSTALL=_build/install/default
 BUILD=_build/default
 SHELL:=/bin/bash
-TIMEOUT=90.0
+TIMEOUT=300.0
```
gares, line 19: Why this?

chetmurthy: The Travis-CI tests were timing out (I think it was the grundlagen test, but I forget); that's why I increased the timeout.  It might be better to change things so that at least for the opam build, only a subset of the tests were run, since elpi is built whenever anything it depends-upon, is changed.

gares: Grundlagen is an important test. On my machine it spends ~ 8 seconds in parsing and about 20 in execution. I'll try it myselft to see if there is a camlp5 regression.

gares: Here I see no difference between master and this PR wrt speed, so please revert this change

chetmurthy: Done.  I will note, that it wasn't for the difference in speed, but rather, so that CI tests wouldn't timeout.

gares:
Thanks for the changes. Still something has to be done w.r.t. the minimal supported version. See https://travis-ci.org/github/LPCIC/elpi/jobs/717650890 for example. Up to now elpi worked on
```
$ export OCAML_MIN=4.04.1
$ export OCAML_MAX=4.09.0
```
I think you may want to bump the max version in the .travis file, but you also need to fix the lower bound I guess. In which version  was `Stdlib` introduced?

chetmurthy:
"minimal supported version"?  I think I'm misunderstanding.  Camlp5 itself works (at least, bootstraps itself) on any Ocaml version >= 4.02 ?  Is there something I broke with my patch?

Uh ... maybe you mean the way I changed Pervasives to Stdlib?  I wasn't sure that that was the right thing to do, and I'm be happy to change 'em back, if you'd prefer *grin*.

gares:
I don't know the Stdlib story. Which range of versions is it supposed to cover? How does camlp5 do?

chetmurthy:
Hmmm ... this link: https://ocaml.org/releases/4.07.0.html
 says that Stdlib was introduced in ocaml version 4.07 in July 2018.  Only two years ago.  I guess, I'd wait  for a few more versions before requiring Stdlib.  There's a package called "stdlib-shims" that, so I have read, can be used to provide Stdlib support for back-level Ocaml versions, but I haven't tried it.

gares:
OK, let's see if CI likes it

gares:
Finally it passes CI.

This PR requires a version of camlp5 which is flagged as "alpha". Is a stable release planned? I don't want this PR to bitrot, but I also don't want to depend on software in alpha.

chetmurthy:
Yes version 8.00 is .... a big, big set of changes, b/c it's bringing camlp5 up-to-date with ocaml's new syntax for PPX rewriting, and a bunch of other syntax changes.  So while the code is pretty stable, there are some minor bugs.  And .... well, I have to do a TON of documentation changes.  But I expect that I should be able to release a stable version sometime in September.  As with this release, I'll make sure that if there are any breaking changes, I send you a PR.  But I don't expect for that to happen.

gares:
Apparently the opam file for elpi was fixes to upper bound the version of camlp5, so no hurry.

gares:
superseeded by #110

PR #82, opened 2020-08-17
Replace sha1sum with shasum in ppxfindcache_aux.ml

MSoegtropIMC:
ppxfindcache_aux.ml uses `sha1sum` which is fairly old and not available on Mac by default (it can be installed with homebrew or Macports). `shasum` is a replacement available on Mac, Linux (tested on Ubuntu 18.04) and Cygwin. Without additional options it should be identical to sha1sum.

### DIFF
--- a/ppxfindcache/ppxfindcache_aux.ml
+++ b/ppxfindcache/ppxfindcache_aux.ml
@@ -63,7 +63,7 @@ let common () =
   let flag = !flag in
   let cache x = List.mem (path_sanitize x) !cache in
 
-  let sha = must @@ exec "sha1sum" [|"-b";file|] in
+  let sha = must @@ exec "shasum" [|"-b";file|] in
   let sha = String.sub sha 0 (String.length sha - 1) in
   let sha = Printf.sprintf "(*%s %s*)\n" sha (String.concat " " (Array.to_list flag)) in
 


gares:
CI fails because of the new camlp5 alpha. There a PR fixing this problem, but I'd rather wait to have a non alpha release. In the meanwhile I should pin camlp5

gares:
I was not planning on doing a release anytime soon.

Can you live with it (for the platform)?
Otherwise I'll make a point release with just this commit atop the last release.

MSoegtropIMC:
@gares : it is no issue to include a patch on top of your releases in opam. But I would prefer to push the patched opam packages to the coq released opam repo (alternatively I can keep it local to Coq platform). I can do the opam PR.

Regarding camlp5: from the beta of the CoqPlatform on all packages will be installed with explicit versions (I don't pin them to not make the live of people harder who want to add packages on top, but the plain platform should always be a consistent set of versions).

gares:
WRT camlp5: The opam file for elpi was fixed, so it should now pull in a compatible version of camlp5. I dunno who did it... but it seems done (I've to backport this change in the opam file at the root of this repo).

gares:
> But I would prefer to push the patched opam packages to the coq released opam repo (alternatively I can keep it local to Coq platform).

OK then, I'll do the point release.

But to be clear: elpi.opam is part of the ocaml standard OPAM repository, while coq-elpi.opam (which requires no fix) is the one in the Coq specific repo. So a new elpi release require no change the coq released opam repo.

MSoegtropIMC:
> But to be clear: elpi.opam is part of the ocaml standard OPAM repository, while coq-elpi.opam (which requires no fix) is the one in the Coq specific repo. So a new elpi release require no change the coq released opam repo.

Sorry, I mixed this up. I meant the opam file should go wherever it belongs - I would prefer to have as few patched opam packages in the coq platform patch repo as possible (because in the end I also want to have a opam meta package for the Coq platform).

MSoegtropIMC:
> OK then, I'll do the point release.

Did you already do the release? Actually I think this is not required - it should be fine to patch this on opam level. I think this would be clearer. The goal of the platform is to include plain releases of components. In this situation I think it would be preferable to use the latest clean release and do the compatibility patch in opam - it is clear for everybody what the change is then. Opam has good facilities for minor patches.

gares:
I plan to do it tomorrow.

gares:
And I take the occasion to also sneak in a few extra fixes and minor improvements.

PR #86, opened 2020-09-07
Remove .install file hack

rgrinberg:
We make the ppx executables private and refer to them by a qualified
path. This allows us to rmeove a hack to edit the .install file after
dune generates it.


gares:
thanks. I guess I must have crewed up something with `project_root` when I tried...

PR #97, opened 2021-04-11
Some more builtins

### LINKED ISSUE #96: Elpi predicate to force garbage collection
I wish to have an Elpi predicate to force a garbage collection. There are some OCaml functions (`Gc.minor`, `Gc.major`, `Gc.full_major`, and `Gc.compact`) to do that and `Gc.compact` is most desirable among those for me. @gares

gares:
Fix #96

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,7 @@
 # UNRELEASED
 
 - API:
+  - New `gc` namespace and type with APIs to access the garbage collector
   - New `min` and `max` operators for the `is` builtin, they work on
     `int` and `real`
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,13 +1,18 @@
 # UNRELEASED
 
 - API:
-  - New `gc` namespace and type with APIs to access the garbage collector
+  - New `gc.control` type for the garbage collector settings
+  - New `gc.get` and `gc.set` for reading and writing GC settings
+  - New `gc.minor`
+  - New `gc.major`
+  - New `gc.full`
+  - New `gc.print_stat`
   - New `min` and `max` operators for the `is` builtin, they work on
-    `int` and `real`
+    `int` and `float`
   - Rename `rex_match` -> `rex.match`
   - Rename `rex_replace` -> `rex.replace`
   - Rename `rex_split` -> `rex.split`
-  - Rename `counter` -> `trace.cuonter`
+  - Rename `counter` -> `trace.counter`
 
 # v1.13.0 (Feb 2021)
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,4 @@
-# UNRELEASED
+# v1.13.1 (April 2021)
 
 - API:
   - New `gc.get` and `gc.set` for reading and writing GC settings
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,19 +1,21 @@
 # UNRELEASED
 
 - API:
-  - New `gc.control` type for the garbage collector settings
   - New `gc.get` and `gc.set` for reading and writing GC settings
   - New `gc.minor`
   - New `gc.major`
   - New `gc.full`
   - New `gc.compact`
-  - New `gc.print-stat`
+  - New `gc.stat`
+  - New `gc.quick-stat`
   - New `min` and `max` operators for the `is` builtin, they work on
     `int` and `float`
   - Rename `rex_match` -> `rex.match`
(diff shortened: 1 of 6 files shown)


CHANGES.md
```diff
@@ -1,3 +1,19 @@
+# UNRELEASED
+
+- API:
+  - New `gc.control` type for the garbage collector settings
+  - New `gc.get` and `gc.set` for reading and writing GC settings
+  - New `gc.minor`
+  - New `gc.major`
+  - New `gc.full`
```
pi8027, line 8: `gc.compact` is missing here.

gares: Thanks for spotting (it is in the new API indeed)

pi8027:
Should they also be exported from `elpi-builtin.elpi` of Coq-Elpi if I wish to use them from Coq-Elpi? It says `constant gc.compact has no declared type` and then `elpi fails`.

gares:
Well spotted. My last commit exports a new list of builtin declarations, which should be passed to API.Setup.init in coq_elpi_vernacular.ml

pi8027 (approved):
By inserting `gc.compact`, I could reduce memory usage. So it seems to work!

pi8027:
Now I think `Gc.stat` and `Gc.quick_stat` are useful to check memory consumption of some computation. Indeed, it is not impossible (intentional double negation ;)) to do so using `Gc.print_stat` but would be easier with `Gc.stat`. Would you mind adding them? (Sorry for asking so many feature requests.)

gares:
What about removing print_stat and keeping only stat and quick_stat ?

pi8027:
> What about removing print_stat and keeping only stat and quick_stat ?

Totally fine with me.

gares:
I've updated the API

pi8027:
I couldn't build Coq-Elpi with this patch and so cannot test:
```
### output ###
# [...]
# CAMLC -c src/coq_elpi_HOAS.ml
# File "src/coq_elpi_HOAS.ml", line 658, characters 33-112:
# 658 |     deprecation = deprecation @@ get_pair_option API.BuiltInData.string API.BuiltInData.string "coq:deprecation";
#                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
# Error: This expression has type
#          (string Elpi.Builtin.unspec * string Elpi.Builtin.unspec) option
#        but an expression was expected of type
#          (string unspec * string unspec) option
#        Type string Elpi.Builtin.unspec is not compatible with type
#          string unspec
# make[1]: *** [Makefile.coq:663: src/coq_elpi_HOAS.cmo] Error 2
# make: *** [Makefile:30: build] Error 2
```

pi8027:
I managed to fix and now I can do this in Coq-Elpi:
```
Elpi Query lp:{{ gc.stat Minor Promoted Major _ _ _ _ _ _ _ _ _ _ _ _ _ }}.
Elpi Query lp:{{ gc.get Minor _ _ _ _ _ _ _ }}.
Elpi Query lp:{{ gc.set 67108864 _ _ _ _ _ _ _ }}.
```

pi8027 (approved):
LGTM

gares:
Oh right, I took the occasion to "move" the unspec data type to elpi, so one has to remove it's copy in coq-elpi

src/builtin.mli
```diff
@@ -54,6 +57,10 @@ val diagnostic : diagnostic API.Conversion.t
 val mkOK : diagnostic
 val mkERROR : string -> diagnostic
 
+type 'a unspec = Given of 'a | Unspec
+val unspec : 'a API.Conversion.t -> 'a unspec API.Conversion.t
+val unspecC : ('a,'b,'c) API.ContextualConversion.t -> ('a unspec,'b,'c) API.ContextualConversion.t
```
pi8027, line 62: I suggest relocating them to `Elpi.API`. (see LPCIC/coq-elpi#230.)

src/builtin.elpi
```diff
@@ -119,6 +119,14 @@ type iabs int -> int.
 
 type rabs float -> float.
 
+type (max) int -> int -> int.
+
+type (max) float -> float -> float.
+
+type (min) int -> int -> int.
+
+type (min) float -> float -> float.
```
pi8027, line 128: I'm not sure whether they should really be infix.

gares: They are not, it's a bug in the pretty printer but has no effect in practice.

PR #107, opened 2021-08-31
unif: fix handling of eta

### LINKED ISSUE #106: `syntax error in as` ?
The following gives me a very strange error (in elpi 1.13.7):

```prolog
type bar any -> any.

foo (bar (uvar as K)) :- print "hello".

main :- (foo (x \ bar u)).
```
outputs
```Fatal error: syntax error in as```

whereas almost any other input:
```prolog
main :- (foo (x \ bar2 u)).
main :- (foo Y).
main :- (foo (bar u)).
main :- (foo (bar X)).
```
succeeds or fails without error.

Is this expected behavior?

gares:
This PR fixed the handling of eta when the term in eta long form is flexible.

closes #106

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,9 @@
+# UNRELEASED
+
+- Runtime/FFI:
+  - fix handling of eta expanded unification variables. Many thanks to
+    @nguermond for testing this tricky case.
+
 # v1.13.7 (July 2021)
 
 - Compiler:
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,7 +2,7 @@
 
 - Runtime/FFI:
   - fix handling of eta expanded unification variables. Many thanks to
-    @nguermond for testing this tricky case.
+    Nathan Guermond for testing this tricky case.
 
 # v1.13.7 (July 2021)
 
--- a/src/runtime.ml
+++ b/src/runtime.ml
@@ -1548,7 +1548,7 @@ let rec restrict_for_eta this xdepth depth adepth e t =
       restrict_for_eta this xdepth depth adepth e
         (deref_appuv ~from:adepth ~to_:depth args e.(i))
   | Lam x ->
-      restrict_for_eta this xdepth (depth+1) adepth e t
+      restrict_for_eta this xdepth (depth+1) adepth e x
   | UVar(r,lvl,ano) ->
       let t, assignment = expand_uv ~depth r ~lvl ~ano in
       option_iter (fun (r,_,assignment) -> r @:= assignment) assignment;
@@ -1558,6 +1558,7 @@ let rec restrict_for_eta this xdepth depth adepth e t =
       option_iter (fun (r,_,assignment) -> r @:= assignment) assignment;
       restrict_for_eta this xdepth (depth+1) adepth e t
   | _ -> true
+
 let restrict_for_eta this xdepth depth adepth e t =
   [%spy "dev:restrict_eta:in" ~rid (fun fmt () ->
       Fmt.fprintf fmt "@[<hov 2>prune %a@ from %a@]"
--- a/src/runtime.ml
+++ b/src/runtime.ml
@@ -1502,63 +1502,63 @@ let occurs x d adepth e t =
   in
   x < d && aux d t
 
-let rec exta_contract_args ~depth r args eat adepth e =
+let rec exta_contract_args ~orig_depth ~depth r args eat adepth e =
   match args, eat with
   | _, [] -> [%spy "eta_contract_flex" ~rid (fun fmt () -> Fmt.fprintf fmt "all eaten") ()];
-      Some (depth,r,List.rev args)
+      begin
+        try Some (AppUVar(r,0,List.map (move ~adepth ~from:depth ~to_:orig_depth e) (List.rev args)))
+        with RestrictionFailure -> None
+      end
   | Const x::xs, y::ys when x == y && not (List.exists (occurs y depth adepth e) xs) ->
       [%spy "eta_contract_flex" ~rid (fun fmt -> Fmt.fprintf fmt "eat %d") y];
-      exta_contract_args ~depth r xs ys adepth e 
+      exta_contract_args ~orig_depth ~depth r xs ys adepth e 
   | _, y::_ ->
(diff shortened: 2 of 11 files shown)


nguermond:
This seems to break the following

```prolog
foo (bar (uvar K Args)) :- print K "@" Args.
main :- (foo (bar (x \ u))).
```
with error
```
Fatal error: as clause too complex
```
in `fix-106`, whereas it failed successfully in `master`.

gares:
should be fixed now

nguermond:
Sorry that I keep finding problems, but this now breaks
```prolog

pred patternify i:A, o:A.

patternify (uvar K Args) U :- print K "@" Args, (distinct_names Args).

%% master: success
%% fix-106: success
main :- patternify X U.

%% master: fails
%% fix-106: Fatal error: (Data.Term.Lam Data.Term.Nil) is not a list
main :- patternify (x \ X) U.

%% master: fails
%% fix-106: Fatal error: (Data.Term.Lam (Data.Term.Cons ((Data.Term.Const c0), Data.Term.Nil))) is not a list
main :- patternify (x \ X x) U.

%% master: fails
%% fix-106: Fatal error: (Data.Term.Lam
%%   (Data.Term.Cons ((Data.Term.Const u),
%%      (Data.Term.Cons ((Data.Term.Const c0), Data.Term.Nil))))) is not a list
main :- patternify (x \ X u x) U.
```

gares:
Thanks for the tests, I will eventually get it right ;-)
I'll try again tomorrow

gares:
This seems a bug in master, no?
```
%% master: fails
%% fix-106: Fatal error: (Data.Term.Lam (Data.Term.Cons ((Data.Term.Const c0), Data.Term.Nil))) is not a list
main :- patternify (x \ X x) U.
```

nguermond:
Yes, I thought so too...

gares:
one more try

nguermond:
One more little snag:
```prolog
bar (uvar as K) :- print K.
bar2 (uvar K) :- print K.

%% master : fails
%% fix-106 : success ??
main :- bar (x \ X).

%% master : fails
%% fix-106 : success ??
main :- bar2 (x \ X).
```

gares:
these are thought, thanks for posting them. I think I'll need yo rethink the code, since here the failures come from the fact that x is not in scope... maybe I should not eta expand but rater eta contract, ir at least fail if I cant eta contract...

gares:
here it is, please test

nguermond:
The following seems to hang:
```prolog
pred foo i:any.
pred foo2 i:any.

foo (uvar K Args).
foo2 (uvar K).

%% main: fails
%% fix-106: does not halt??
main :- foo (x \ y \ X x y).
main :- foo2 (x \ y \ X x y).

%% main: fails
%% fix-106: fails
main :- foo (x y \ X x y).
main :- foo2 (x y \ X x y).
```
Should these succeed?
Otherwise this seems to work well for all practical purposes.

gares:
Here the terms are really `x` applied to `y\ X x y`. While `pi` and `sigma` support multiple binders at once, lambda alone can't
```
main :- foo (x y \ X x y).
main :- foo2 (x y \ X x y).
```

gares:
The code I wrote sees `(x \ y \ X x y)` as the eta expansion of `X` so it passes. Do you find this surprising?

nguermond:
What I mean is that `foo (x \ y \ X x y)` does not terminate for me.

gares:
Right, sorry it was a very silly typo

nguermond:
It seems `X` is substituted for `c \ X1` rather than `c \ X1 c`:
```prolog
pred foo i:any.

foo X :- print X, fail.
foo (uvar as X) :- print X.

main :- (foo (x \ X x)),
		print X.
```
outputs
```
c0 \ X0 c0
X1
c0 \ X1
```

gares:
Many thanks for your help!
I did rewrite the whole code. It is more correct I believe and passes these tests, but there is also a bug I won't be able to fix. In particular occur-check is not up to eta, that is `(x\ X x) = X` fails.

The `as` clause has changed a little (w.r.t. the previous attempts) so that if you name a term like `x\X x`, you get that term, even if you write `uvar as T`. I think this is better, since `as` and `eta` are not orthogonal. Then the `var` builtin was also fixed to eta-contract what you give, so a rule like `foo (uvar as T) :- var T, print T.` works on `foo (x\ X x)` but prints `c0\ X c0` unlike in previous commits.

nguermond:
This doesn't seem so bad compared to what we get!

There still seems to be a little problem with
`main :- pi x \ (var (y \ X x y)).` failing whereas both
```
main :- pi x \ (var (y \ X c y)). 
main :- pi x z \ (var (y \ X x y)).
```
succeed (but this works as expected with `uvar`/`as`).

Another minor remark: should `declare_constraint` also accept eta-expanded variables?

gares:
Thanks again for your tests. I did fix this glitch and also called eta_contract_flex in declare_constraint

CHANGES.md
```diff
@@ -1,3 +1,9 @@
+# UNRELEASED
+
+- Runtime/FFI:
+  - fix handling of eta expanded unification variables. Many thanks to
+    @nguermond for testing this tricky case.
```
gares, line 5: @nguermond I can put your name if you prefer.

src/runtime.ml
```diff
@@ -1467,6 +1469,97 @@ let error_msg_hard_unif a b =
...
+    | AppArg(i,args) when e.(i) != C.dummy ->
+      aux d (deref_appuv ~from:adepth ~to_:d args e.(i))
+    | Nil | CData _ | Discard | Arg _ | AppArg _ -> false
+  and auxs d = function
+    | [] -> false
+    | t :: ts -> aux d t || auxs d ts
+  in
+  x < d && aux d t
+
+let rec exta_contract_args ~orig_depth ~depth r args eat adepth e =
```
nguermond, line 1505: is exta a typo?

nguermond:
This seems good to me, thank you so much for all your work!

CHANGES.md
nguermond: It's Nathan Guermond, thanks!

src/runtime.ml
gares: thanks

PR #112, opened 2021-09-09
Improve init docs

shonfeder:
Points to Elpi.Builtin instead of Elpi_builtin, and uses some document
syntax to pretty up the documentation.

Here's what the docs look like rendered with these changes: 

2021-09-08-202447_1492x958_scrot

I'm not 100% the use of params is an improvement. Let me know if you'd rather I drop that and just fix the module reference.

I hope this superficial fix is not too presumptuous :)

I am so glad this library and interpreter exists :)

### DIFF
--- a/src/API.mli
+++ b/src/API.mli
@@ -44,12 +44,15 @@ module Setup : sig
   type elpi
 
   (** Initialize ELPI.
+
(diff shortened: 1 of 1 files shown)


gares (approved):
It's great, thanks!

PR #119, opened 2021-11-09
Improve behavior and error messages in nth

mb64:
Currently, `nth` issues a single error message `nth run out of list items`, for a few situations:
 - when the index is too large;
 - when the index is negative;
 - and when it successfully indexes the collection, but unifying this element with the result fails. (This is due to improper cut placement.)

This change keeps the error message for the first situation, but adds a `nth got a negative index` error message for the second, and makes it fail normally in the third case.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,8 @@
     Nathan Guermond for testing this tricky case.
   - Change `Rawdata.Constants.eqc` to a builtin
   - Fix `Rawdata.Constants.cutc` has always been a builtin
+- Library:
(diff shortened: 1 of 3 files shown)


gares:
great, thanks!
But the file you should change is https://github.com/LPCIC/elpi/blob/master/src/builtin_stdlib.elpi and the run make build to update the other.
You may also add a changelog line about this fix.

mb64:
Thanks, fixed.

gares (approved):
thanks

gares:
great, running ci then merging!

Issue #139, opened 2022-03-27
Elpi 1.15.0 doesn't compile with released OCaml 4.14.0 if camlp5 is installed 


Blaisorblade:
Even if OCaml 4.14 is unreleased, RC2 is out and @rlepigre was interested in upgrading to the upcoming release, so I tried building the latest Elpi with trunk, and it fails with this error log https://gist.github.com/Blaisorblade/3bfadf8bbf45030dac539211f8e188ca.

This seems the most interesting part:

```
File "src/parser.ml", line 140, characters 36-53:
140 |    let res = Grammar.Entry.parse e (Stream.of_channel ch)in
                                          ^^^^^^^^^^^^^^^^^
Alert deprecated: module Stdlib.Stream
Use the camlp-streams library instead.
File "src/parser.ml", line 140, characters 35-57:
140 |    let res = Grammar.Entry.parse e (Stream.of_channel ch)in
                                         ^^^^^^^^^^^^^^^^^^^^^^
Error: This expression has type char Stdlib.Stream.t
       but an expression was expected of type char Stream/1.t
       Stream/1.t is abstract because no corresponding cmi file was found in path.
```

rlepigre:
This is due to https://github.com/ocaml/ocaml/pull/10896, I think this should be solved by adding `camlp-streams` as dependency.

gares:
This comes from camlp5. would you mind trying the `adt/parser` branch in a switch where camlp5 is not installed? You then need https://github.com/LPCIC/coq-elpi/pull/350 .

If this work, I won't do anything about this issue, release elpi 1.15 ASAP and merge that pr on coq-elpi. If it does not work, I'll investigate this API change.

gares:
Ah ok, if the fix is that simple, I can apply it.

rlepigre:
Branch `adt/parser` seems to compile/install fine on `4.14.0+trunk` (with a few deprecation warnings).

Blaisorblade:
The same problem seems to apply to elpi.1.15.0 and the OCaml 4.14 release — am I doing something wrong?
```
$ opam install elpi.1.15.0
The following actions will be performed:
  ∗ install menhirSdk 20220210 [required by menhir]
  ∗ install menhirLib 20220210 [required by menhir]
  ∗ install menhir    20220210 [required by elpi]
  ∗ install elpi      1.15.0
===== ∗ 4 =====
Do you want to continue? [Y/n] y

[...]

PR #140, opened 2022-03-27
Fix compilation on OCaml 4.14.

rlepigre:
This addresses https://github.com/LPCIC/elpi/issues/139.

### DIFF
--- a/elpi.opam
+++ b/elpi.opam
@@ -14,7 +14,7 @@ build: [
 ]
 
 depends: [
-  "ocaml" {>= "4.04.0" & < "4.15.0"}
+  "ocaml" {>= "4.04.0"}
   "stdlib-shims"
   "camlp-streams"
   "camlp5" {> "7.99"}
--- a/elpi.opam
+++ b/elpi.opam
@@ -14,8 +14,9 @@ build: [
 ]
 
 depends: [
-  "ocaml" {>= "4.04.0" & < "4.14.0"}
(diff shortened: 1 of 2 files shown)


elpi.opam
```diff
@@ -14,8 +14,9 @@ build: [
 ]
 
 depends: [
-  "ocaml" {>= "4.04.0" & < "4.14.0"}
+  "ocaml" {>= "4.04.0" & < "4.15.0"}
```
gares, line 17:
```suggestion
  "ocaml" {>= "4.04.0"}
```

gares:
thanks

gares:
Something very fishy ;-)
```
Error: This expression has type char Stream.t
but an expression was expected of type char Stream.t
```

rlepigre:
> Something very fishy ;-)
> 
> ```
> Error: This expression has type char Stream.t
> but an expression was expected of type char Stream.t
> ```

Oh my... That's annoying. I guess that's what you get when there are two `Stream` modules around. :)
No idea how to fix that though, `camlp5` is probably using the module from the standard library.

I guess the point of https://github.com/LPCIC/elpi/pull/130 is to avoid doing this altogether, right? In that case, I guess we can just wait for https://github.com/LPCIC/elpi/pull/130, since it seems you already made good progress there @gares. ;)

rlepigre:
(Perhaps we could try to tweak the linking order, but that would probably be fragile.)

PR #146, opened 2022-05-09
Trace elaborator

gares:
TODO:
- [x] depend on atd >= 2.9.1


gares:
TODO: resumed goal should contain the updated text, not the original one

PR #155, opened 2022-07-01
Trace fix failure

gares:


gares:
@cohencyril @jwintz this hopefully fixes the bug making the trace browser fail on a "broken" trace, which is the result of a coq command which fails abruptly or an elpi program which bails out via a fatal error. I plan to make a new elpi release with it today.

jwintz:
Would it be possible to commit:
- `broken_trace1.elpi`
- `broken_trace2.elpi`

In addition to manual testing, these could also be handled by the documentation's injection engine for integration ...

gares:
the former one cmes from a real failure in HB which is a coq-elpi app, so you need coq. The second trace is fabricated from the former.

So no, sorry

PR #157, opened 2022-07-16
elpi.opam: depend on OCaml 4.08

Blaisorblade:
(Seems) needed to match the changelogs in https://github.com/LPCIC/elpi/releases, assuming those are correct — haven't checked.


gares (approved):

gares:
thanks

PR #160, opened 2022-09-01
Hide unsafe code from the optimiser

lthls:
The type definition for `alternatives` is a record, so a clever compiler (such as Flambda 2) can assume that values of this type can never be equal to `0`. This translates all `alts == noalts` checks into trivial `false` constants, eventually breaking down with a segfault when `noalts` is actually passed.
My proposed fix is the minimal one: hide the constant `0` from the compiler, so it can't optimise away.
The real fix is to use a variant definition. Example:
```ocaml
and alternative =
| Noalts
| Alts of {
  cutto_alts : alternative;
  program : prolog_prog;
  adepth : int;
  agoal_hd : constant;
  ogoal_arg : term;
  ogoal_args : term list;
  agid : UUID.t; [@trace]
  goals : goal list;
  stack : frame;
  trail : T.trail;
  state : State.t;
  clauses : clause list;
  next : alternative;
}
```
It's exactly the same runtime representation as the current one, but type-safe.

### DIFF
--- a/elpi.opam
+++ b/elpi.opam
@@ -25,8 +25,8 @@ depends: [
   "cmdliner" {with-test}
   "dune" {>= "2.8.0"}
   "conf-time" {with-test}
-  "atdgen" {>= "2.9.1"}
-  "atdts" {>= "2.9.1"}
+  "atdgen" {>= "2.9.1" & < "2.10.0"}
+  "atdts" {>= "2.9.1" & < "2.10.0"}
 ]
 depopts: [
   "elpi-option-legacy-parser"
--- a/src/runtime.ml
+++ b/src/runtime.ml
@@ -2842,7 +2842,7 @@ and alternative = {
   clauses : clause list;
   next : alternative;
 }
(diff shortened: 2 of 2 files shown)


gares:
I've already undone this optimization in #118, that that PR may need more months to complete. 
Is Flambda 2 out?

lthls:
No, Flambda 2 is not out yet, but it's in a good enough state that we're doing opam-wide testing. We can deal with packages failing with known issues, so it's not particularly urgent that you fix this issue.
If you plan a release before your other PR is merged, then including this patch would be nice as it would make our CI happy again, otherwise you can just ignore this.
We might actually fix it on our side by tweaking the implementation of `Obj.magic` anyway.

gares:
CI fails because I did not port here a fix to the opam package which was done upstream.
I'm fine merging this, but I lack time, so if you could update the opam file here and make CI happy, that would be great.
I shall release before tabling is merged.

gares:
Thanks!

Issue #159, opened 2022-08-15
Compilation issue with yojson 2.x.x


SnarkBoojum:
If one compiles a more recent yojson (2.0.2 here), then use it against ocaml-atd 2.9.1, then the resulting lib breaks elpi:
```
File "src/elpi_trace_elaborator.ml", line 668, characters 14-16:
668 |   write_trace ob cards;
                    ^^
Error: This expression has type Bi_outbuf.t
       but an expression was expected of type Buffer.t
```
because indeed in the 2.x.x series, yojson doesn't use biniou anymore, so Bi_outbuf isn't there anymore.

I noticed it because I'm looking into updating yojson in Debian, so I could find everything that breaks and report to the various upstreams.

I'll try to provide a patch.

xgqt:
I just ran into this issue on Gentoo.

@SnarkBoojum have you had time to see if this can be patched?

SnarkBoojum:
I have no clue where `write_trace` comes from... the three lines can probably replaced with something like `List.map (fun card -> Yojson.Safe.to_channel stdout card) cards` ; but first cards needs to be turned into `Yojson.Safe.t` and I don't know how.

SnarkBoojum:
(now that I think of it, `write_trace` is probably from trace.atd)

gares:
> some output is different so it breaks some tests

how? (not that I did a release on Friday since the test suite was broken)

gares:
Oh, but this change is more about `ydump` than your patch. Maybe it has a flag to stick to the old format?

gares:
(yojson prints a single line, I then run ydump on it to make the test reference readable)

SnarkBoojum:
It's not backward-compatible ; in fact I'm using yojson 2.0.2 and atd 2.10.0.

SnarkBoojum:
I added a commit to tighten the dependency on atd.

PR #167, opened 2022-10-24
Fix for yojson 2.0.2

SnarkBoojum:
This is a fix for issue #159.


gares:
Thanks, but please also relax the constraint in `elpi.opam` (in another commit, I want this CI to run anyway)

gares:
So the new code is not backward compatible.

gares:
thanks

PR #173, opened 2023-02-09
Open Gramlib in legacy parser

gridbugs:
This is needed to bring `Ploc` into scope.

### DIFF
--- a/src/legacy_parser/parser.ml
+++ b/src/legacy_parser/parser.ml
@@ -9,6 +9,7 @@ module Loc = U.Loc
 open Elpi_parser.Ast
 open Elpi_parser.Parser_config
 open Term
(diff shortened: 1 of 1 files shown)


gares:
Hum, ci is not happy. Which setup do you have?

gridbugs:
Interesting. I've been testing on ubuntu with ocaml 4.14.1. I'll take a look at the failure today.

gridbugs:
This is likely due to me vendoring the source code for camlp5 into my project and hacking it to build with dune. The `Gramlib` module is from the `camlp5.gramlib` library. I suspect I'm missing a `(wrapped false)` somewhere. This is not a problem with elpi. Closing.

PR #174, opened 2023-02-19
Run dune build and check in modified files

smuenzel:
After running `dune build`, the workspace isn't clean. Fix this by running `dune build` and commiting.


gares:
This is probably due to a newer (or older) atd version. Which version are you using?
I'm ok merging if it is the latest one.

smuenzel:
Ah I see. I was using 2.10.0, but 2.11.0 also produces the diff in this patch. I believe that's the latest one according to opam

gares:
thanks!

PR #184, opened 2023-06-03
improve separate compilation API performance

gares:
todo:
- [x] why spill run on the whole program (and not just the new part)?
- [x] merge_types is quadratic 
- [ ] why insertion on the whole list, and not per-predicate?
- [x] why clauses are kept in regular order is we typically append to them? (keep them in rev order)

### DIFF
--- a/src/compiler.ml
+++ b/src/compiler.ml
@@ -435,7 +435,7 @@ type program = {
   toplevel_macros : macro_declaration;
 }
 and pbody = {
-  types : typ list C.Map.t; (* TODO: use a map : symbol -> types to speed up merge types *)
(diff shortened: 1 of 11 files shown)


src/compiler.ml
```diff
@@ -435,15 +435,15 @@ type program = {
   toplevel_macros : macro_declaration;
 }
 and pbody = {
-  types : typ list;
+  types : typ list C.Map.t; (* TODO: use a map : symbol -> types to speed up merge types *)
```
gares, line 438:
```suggestion
  types : typ list C.Map.t;
```

PR #210, opened 2023-11-29
HO api

gares:
Fix LPCIC/coq-elpi#549


MSoegtropIMC:
Many thanks! I didn't imagine that it would become so complicated ...

PR #214, opened 2023-12-07
Discr tree multivar

FissoreD:
Goal:
if `L = [1,2,3]` is a list, then the corresponding path used for indexing with discrimination tree is 
`[Cons_2, Primitive, Cons_2, Primitive, Cons_2, Primitive, Nil_0]` where `Cons_2` (resp `Nil_0`) tells that we
are dealing with a constant of name `Cons` (resp. `Nil`) of arity `2` (resp. `0`)
In this encoding in each list, half of the size of the path is used for `Cons_2` and `Nil_0`. 
With this PR, we reduce the size of the path by adding two special marker for the Head and the End of a list and
remove all the Cons_2 and Nil_0.
The path for the list L will be encoded as `[ListHead, Primitive, Primitive, Primitive, ListEnd]`


src/discrimination_tree.mli
```diff
@@ -0,0 +1,63 @@
...
+val isListTailVariable : cell -> bool
+val isPathEnd : cell -> bool
+
+module Trie : sig
+  type 'a t
+  val pp : (Format.formatter -> 'a -> unit) -> Format.formatter -> 'a t -> unit
+  val show : (Format.formatter -> 'a -> unit) -> 'a t -> string
+end
+
+type 'a data = { data : 'a; time : cell; }
```
gares, line 30: This could be opaque

src/discrimination_tree.mli
```diff
@@ -0,0 +1,63 @@
...
+type 'a t = { t : 'a data Trie.t; max_size : int; max_depths : int array; }
+
+val index : 'a t -> path -> 'a -> time:int -> 'a t
+val max_path : 'a t -> int
+val max_depths : 'a t -> int array
+
+val empty_dt : 'b list -> 'a t
+
+val retrieve : path -> 'a t -> 'a list
```
gares, line 39: This should document the Order of the list

src/discrimination_tree.mli
```diff
@@ -0,0 +1,63 @@
...
+val isVariable : cell -> bool
+val isLam : cell -> bool
+val isInput : cell -> bool
+val isOutput : cell -> bool
+val isListHead : cell -> bool
+val isListEnd : cell -> bool
+val isListTailVariable : cell -> bool
+val isPathEnd : cell -> bool
+
+module Trie : sig
```
gares, line 24: This could be hidden, or go intorno the invernale module

gares:
It seems the branch needs to be rebased

PR #242, opened 2024-07-10
Fix loc after file resolution

gares:
@FissoreD would you mind testing this version of Elpi? It should give (in coq-elpi) better file names in error messages


FissoreD:
Cannot find the branch:
image

FissoreD:
What command should I write to get it, I did git remote update, but still not found

gares:
`opam pin add elpi https://gihub.com/LPCIC/elpi.git#fix-loc-after-file-resolution`

gares:
I like the approach, **** dune!

gares:
elpi already depends on `Re`, and `open Re` gives you a `Str` module.
you should remove the `str` dependency and just open `Re` where needed.

src/parser/parse.ml
```diff
@@ -67,6 +67,7 @@ let already_parsed = Hashtbl.create 11
 let () =
   parse_ref := (fun ?cwd filename ->
   let filename = C.resolver ?cwd ~unit:filename () in
+  let dest = Re.Str.replace_first (Re.Str.regexp "/_build/default") "" filename in
```
gares, line 70:
```suggestion
  let dest = Re.Str.replace_first (Re.Str.regexp "/_build/[^/]+") "" filename in
```

PR #246, opened 2024-07-23
CHR: constraints allow hyps, not considered in the clique error check

FissoreD:
The grammar has been extended to support the new behavior

A test has been added


src/compiler.ml
```diff
@@ -460,7 +460,7 @@ and block =
   | Clauses of (preterm,Ast.Structured.attribute) Ast.Clause.t list (* TODO: use a map : predicate -> clause list to speed up insertion *)
   | Namespace of string * pbody
   | Shorten of C.t Ast.Structured.shorthand list * pbody
-  | Constraints of constant list * prechr_rule list * pbody
+  | Constraints of constant list * constant list * prechr_rule list * pbody
```
gares, line 463:
```suggestion
  | Constraints of constraints_block * pbody
 and constraints_block = {
   clique : constant list;
   ctx_filter : constant list;
   rules : prechr_rule list;
 }
```

and use the type also below instead of the triple

gares (changes requested):
Please define a type with names so that there is no risk of confusing the hyps with the rest. Also, I think `?-` is a better symbol than `|-` since it is the same.

gares:
There is something I don't get. `CHR.new_clique` should now have a different type, a clique should be two sets on symbols, not the union of the two sets. And the code filtering the context should use only that set of predicate names.

gares:
Also, I expect to see some code making the union of all the context filters (for the same clique).

gares:
https://github.com/LPCIC/elpi/blob/af3bb10e6aa447a9c968227432b0bd0d75c439ca/src/runtime.ml#L3417-L3422

gares:
`clique_of` should also return the context filter, I guess

FissoreD:
Yeah, I'm going in that direction:
`type clique = {ctx_filter: Constants.Set.t; clique: Constants.Set.t} [@@deriving show]`

FissoreD:
I'm however a bit lost on how to correctly manage the ctx_filter

gares (approved):

gares:
please rebase and update the README.md

ELPI.md
```diff
@@ -461,15 +461,62 @@ The `constraint` directive gives control on the hypothetical part of the
...
+}
+```
+
+> [!IMPORTANT]
+> When a suspended goal (via `declare_constraint`) is resumed, only the rules
+implementing the symbols passed in the head of the constraint are kept. 
+
+In our example, only the rules for `p1, p2, p3, foo` and `bar` are kept.
+Therefore, if just before the suspension of the goal `foo x` a rule like `p4`
+exists, this rule will not be filtered out from the rules in the suspended goal.
```
gares, line 477:
```suggestion
exists, this rule will be filtered out from the context of the suspended goal.
```

ELPI.md
```diff
@@ -461,15 +461,62 @@ The `constraint` directive gives control on the hypothetical part of the
...
+
+The symbol `?-` in the head of a constraint is used to separate two lists of
+predicate names, the former list represents predicate names to be loaded as a
+predicate filter to load the wanted rules as hypotheses in the context, whereas
+the latter list contains predicates to be used in the new premises to be solved.
```
gares, line 482:
```suggestion
The symbol `?-` separates two lists of
predicate names: the former list is a predicate filter for the context;
the latter list is a predicate filer for the goal.
```

ELPI.md
```diff
@@ -461,15 +461,62 @@ The `constraint` directive gives control on the hypothetical part of the
...
+
+In the example above, the rules implementing `p1`, `p2` and `p3` are filtered in
+in the suspended goal, that is, they are loaded into `Ctx`. Therefore, when
+solving the goal  `Ctx => bar Arg1` all the rules for these three predicates are
+charged.
```
gares, line 487:
```suggestion
In the example above, the rules implementing `p1`, `p2` and `p3` are kep in
in the suspended goal context. Therefore, when
solving the goal  `Ctx => bar Arg1` all the rules for these three predicates are
part of `Ctx`.
```

ELPI.md
```diff
@@ -461,15 +461,62 @@ The `constraint` directive gives control on the hypothetical part of the
...
+
+The list of predicate names after the `?-` should form a "clique", a set of
+symbols disjoint from all the other cliques in the constraint store. If two
+overlapping cliques are detected, the fatal error *overlapping constraint
+cliques* is raised. The overlapping check is not applied to the list of
+hypotheses symbols before `?-`, that is, in the case of two constraints declared
+on two same cliques `c` with different hypothesis `h1` and `h2`, then the two
+set of hypotheses are merged and added to the clique `c`.
```
gares, line 495:
```suggestion
cliques* is raised. The overlapping check is not applied to the context filters,
that is, in the case of two constraints declared
on two same cliques `c` with different filters `h1` and `h2`, then the two
filters are merged and added to the clique `c`.
```

ELPI.md
```diff
@@ -461,15 +461,62 @@ The `constraint` directive gives control on the hypothetical part of the
...
   rule ...
 }
 ```
-The effect is that whenever a goal about `foo` or `bar`
-is suspended (via `declare_constraint`) only its hypothetical
-clauses about `foo` or `bar` are kept.
+
+From now on, all the goals suspended on the predicates `foo` and `bar` will see
+in their contexts the four all the rules implementing the predicates `p1, p2,
+p3, p4` = $\{$`p1,p2,p3`$\} \cup \{$`p4`$\}$.
```
gares, line 508:
```suggestion
in their contexts the all the rules implementing the predicates `p1, p2,
p3, p4` = $\{$`p1,p2,p3`$\} \cup \{$`p4`$\}$.
```

ELPI.md
```diff
@@ -461,15 +461,62 @@ The `constraint` directive gives control on the hypothetical part of the
...
+
+> [!NOTE]
+> If the list of predicate names before `?-` is empty, then the `?-` can be avoided
```
gares, line 511:
```suggestion
> If the list of predicate names before `?-` is empty, then the `?-` can be omitted  
```

gares (approved):

PR #248, opened 2024-07-24
Store symbol tables in a more compact way for compiled programs.

ppedrot:
While not crazy, with this trivial patch I get a non-negligible size reduction of `mathcomp/ssreflect/order.vo` from 11817893 bytes to 11087515 bytes (~6%).


ppedrot:
We can squeeze the data a bit more, I'll submit another PR after this one.

gares (approved):

PR #249, opened 2024-07-25
More compact representation of compiled programs

ppedrot:
Follow-up of #248, basically made of two units.

- Make the representation of the pruned symbol more compact by dropping useless data and representing the map as an array.
- Reduce the size of the Flat.program type by hoisting out data that is actually not semantically part of this record.

This reduces the size of order.vo from 11087515 bytes to 10349640 bytes, i.e. another compounded 6%.

EDIT: the last commit compacts this file even more, to 10062305 bytes.

### DIFF
--- a/src/compiler.ml
+++ b/src/compiler.ml
@@ -546,8 +546,6 @@ type program = {
   clauses : (preterm,Ast.Structured.attribute) Ast.Clause.t list;
   chr : (constant list * prechr_rule list) list;
   local_names : int;
-
-  toplevel_macros : macro_declaration;
 }
 [@@deriving show]
 
@@ -591,7 +589,7 @@ type compilation_unit = {
 
 type builtins = string * Data.BuiltInPredicate.declaration list
 
-type header = State.t * compilation_unit
+type header = State.t * compilation_unit * macro_declaration
 type program = State.t * Assembled.program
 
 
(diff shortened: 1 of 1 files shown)


src/compiler.ml
```diff
@@ -105,27 +105,37 @@ type table = {
   uuid : Util.UUID.t;
 } [@@deriving show]
 
-type pruned_table = {
-  c2s0 : string D.Constants.Map.t;
-  c2t0 : D.term D.Constants.Map.t;
-} [@@deriving show]
+type entry =
+| ENeg of D.constant * string
+| EPos of D.constant * D.term
```
gares, line 110: ENEg should be something like GlobalSymbol (* < 0 *)
EPos should be BoundVariable (* >= 0 *)

gares: I'm a bit surprise it is not 2 arrays, IIRC the bound variable one is small and the other is contiguous, so using the abs of the value as the array position seems more natural/understandable and would remove half of the data in the buckets

gares: I mean, D.constant is an int, -1 ... -maxg for globals, 0 .. maxl for locals, where maxl is typically not more than 3, while maxg can be large (thousands)

ppedrot: I went for the most straightforward changes, we can try to go even further. I don't think that splitting the array will matter much though.

> the other is contiguous

Is it still the case after pruning? The set of live symbols could also contain globals IIUC.

ppedrot: (I did change the constructor names for clarity.)

gares (approved):

Issue #256, opened 2024-08-02
Unification in the head of a rule and move of binders


FissoreD:
```prolog
pred u o:(int -> int).
u (y\y).

pred main.
main :-
  pi x\ u (y\ X x y), print "The result X" (X x).
```

In the code above, the code should unify `(y\ X x y)` with `(y\y)`.
However, the result printed by `print "The result X" (X x)` is `y\x`.

I've inspected the trace and see that at a certain moment, the interpreter does the following unification:
```
^1:c1 \ .X0 c0 c1 =0= ^0:c0 \ c0
then: 
^1:.X0 c0 c1 =1= ^0:c0 
then:
move c0 from 0 to 1
then:
^1:.X0 c0 c1 =1= ^1:c0 
```
this last unification succeeds and assigns `x\y\x` to `X0`

I think the term to be unified should be `^1: X0 yc0 c1 =0= ^1:c1` (the right-hand side is c1 and not c0), but don't know 
where to change the code...

As a small remark, note that the very similar program below behaves as I expect:
```prolog
pred main.
main :-
  pi x\ (y\ X x y) = (y\y), print "The result X" (X x).
```

FissoreD:
Maybe this script is more similar to the one using the `u` predicate:
```prolog
pred main.
main :-
  Z = (y\y),
  pi x\ (y\ X x y) = Z, print "The result X" (X).
```

gares:
@FissoreD please have a look at #259

PR #257, opened 2024-08-05
unification of (lam t1) (lam t2) moves t1 and t2 at same level + test

FissoreD:
Attempt to solve #256:

When unifying two terms t1 and t2 starting with the lam constructor.
Let d1 (resp d2) the level of t1 (resp t2).
Before unifying the body of t1 and t2, we want to put the two terms at the same level.

> [!WARNING]  
> To make two terms at the same level, a call to move is done. This means that 
> if the term is not in the PF the error _Non deterministic pruning, delay to be implemented_ is raised

For example, the term `pi dummy\ clause2 (x\y\x,x)` raise the error when trying to unify with 
the rule `clause2 (x\y\X x, F y (X x)) :- F = a\b\b.`
This is due to the call to move is done before the assignement of `X` to the identity function.
This means that no deref can be done on `X` and `F y (X x)` which is not in $\mathcal{L}_\lambda$

### DIFF
--- a/src/runtime.ml
+++ b/src/runtime.ml
@@ -1636,7 +1636,6 @@ let rec unif argsdepth matching depth adepth a bdepth b e =
        bdepth (ppterm (bdepth+depth) [] ~argsdepth e) b)
    begin
    let delta = adepth - bdepth in
-         
    (delta = 0 && a == b) || match a,b with
     | (Discard, _ | _, Discard) -> true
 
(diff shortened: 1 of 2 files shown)


gares:
It may fix the problem but it is not a desired fix. The unification works on terms that are not at the same level exactly to avoid the costly move... So IMO the bug occurs after, when the terms under the lambda are unified. Sorry but I don't have the time to look into it right now, but you can work with this patch temporarily

FissoreD:
No problem, we will discuss about it next month. Have a nice holiday ! :wave:

gares:
Closing this one since the bug is fixed

PR #264, opened 2024-09-10
Compile clause early

gares:
@FissoreD it is a bit too untested, but should remove some overhead in `optimize`.
If you have the bench infrastructure for Coq-Elpi at hand, it would be nice to try to out.


gares:
in favor of #265

PR #266, opened 2024-09-29
[typeabbrev] type `dl (dl A)` where `dl` is typeabbrev is no more "looping"

FissoreD:
```prolog
typeabbrev (dl A) (list A).
pred p i:dl (dl A).
```
`dl (dl A)` is no more considered as an invalid (looping) type.

A typeabbrev can be created only if all of the constants in it have already been declared before it.
Recursive typeabbrev (like `typeabbrev dl dl` are still forbidden)

### DIFF
--- a/src/compiler.ml
+++ b/src/compiler.ml
@@ -430,6 +430,7 @@ type type_abbrev_declaration = {
   tavalue : preterm;
   taparams : int;
   taloc : Loc.t;
+  timestamp:int
 }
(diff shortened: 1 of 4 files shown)


gares:
Can't we detect statically that the typeabbrev is not silly, as in the LHS does not occur in the RHS (and the RHS is only made of known symbols)? The last part may require some cleanup (planned but not done).

Issue #272, opened 2024-10-22
Apparent problem with indexing via discrimination trees


swasey:
The following fails with a timeout on my system.[^bluerock]

Dropping the `:index(1 1)` on `std.map2` makes it go through.

I noticed the problem via `#[only(tag)] derive` for a type with more than 32 constructors. In that context, `map2` immediately failed (w/o invoking its callback).

[^bluerock]: BlueRock uses some minor patches to Elpi starting from 0118abf. I'll confirm with elpi, coq-elpi master soon.

```
Require Import elpi.elpi.

Elpi Program test lp:{{
	pred any_list i:int, o:list int.
	any_list 0 [].
	any_list N [N|L] :- any_list {calc (N - 1)} L.

	pred any_pred i:int, i:int, o:int.
	any_pred A B R :- R is A + B.

	pred test i:int, o:list int.
	test N R :- any_list N L, std.map2 L L any_pred R.
}}.
Elpi Typecheck.

Succeed Timeout 10 Elpi Query lp:{{ test 33 L }}.
```

swasey:
> I'll confirm with elpi, coq-elpi master soon.

@Janno confirmed the timeout with elpi 1.20.0 and coq-elpi master.

gares:
Thanks for reporting. @FissoreD would you mind looking into this bug when you have the time?

gares:
33 loops, 32 works...
The bug is related to this piece of code:
```
  let rec list_to_trie_path ~safe ~depth ?(h=0) path_depth (len: int) (t: term) : unit =
    match deref_head ~depth t with
    | Nil -> Path.emit path mkListEnd; update_current_min_depth path_depth
    | Cons (a, b) ->
        (* heuristic: we limit the size of the list to at most 30 *)
        (*            this aims to control the width of the path, *)
        (*            or equivalently the number of children of   *)
        (*            a specific node.                            *)
        (* for example, the term `app[1,...,100]` with depth 2,   *)
        (*              has the node `app` with arity `1` as first*)
        (*              cell, then come the elment of the list    *)
        (*              up to the 30^th elemebt                   *)
        if h > 30 then (Path.emit path mkListEnd; update_current_min_depth path_depth)
        else
          main ~safe ~depth a path_depth;
          list_to_trie_path ~depth ~safe ~h:(h+1) path_depth (len+1) b
```
[...]

PR #274, opened 2024-10-23
[dt] #272 rebased

FissoreD:
This solves bug from issue #272 (parentheses forgotten in the code)
Moreover, it reduces the size of paths: for each predicate we store the size of the longest list in its rules and use this length to prevent lists in the goal to be longer then this bound


gares (approved):
A comment describing the difference between listTailVar and listTailVarUnif would help, otherwise it looks fine to me.

FissoreD:
I think everything is OK, I just want to add the comment

PR #278, opened 2024-10-28
[builtins] std.list.init and std.list.make

FissoreD:


gares:
Most of the list related APIs are in `std.`, possibly by mistake.
I think `std.list.init` and `std.list.make` are more appropriate names for these two.

FissoreD:
I think the same.
I prefer std.list.null rather then std.null
The same for map, map2, ...

FissoreD:
Should we put the whole think in a namespace like sets and maps ?

gares:
that would break a lot of code, so at least we should keep (deprecated?) aliases. Maybe not in this pr

gares (approved):

PR #279, opened 2024-10-29
Scope term add impl constructor

FissoreD:


FissoreD:
It seems a problem from the base branch.
I have promoted the jsons without looking to their result
But in fact `dune exec elpi -- -test tests/sources/printer.elpi` has a wrong printing
I think it is due to the checkbox not marked as done [here](https://github.com/LPCIC/elpi/pull/269#issue-2570851478)

PR #280, opened 2024-10-30
[functionality] add functionality field to program

FissoreD:
Yet no functionality check but we have the infrastructure to know what predicates are functional and what is the functionality relation of its arguments.

~~This partially works with type abbreviations, a wrong compiled functionality is given by the following program:~~
```
typeabbrev (x A) (:functional pred i:A).
:functional pred p i:(x (:functional pred i:int)).
```
~~where the functionality of the argument `A` of `x` is forgotten~~

~~To solve this issue we need an abstraction constructor in the functionality type~~ $\to$ done


src/compiler.ml
```diff
@@ -3899,17 +4024,18 @@ end = struct
...
-    if hash = base_hash then
-      precomputed_kinds, precomputed_type_abbrevs, precomputed_types
-    else
-      let kinds = Flatten.merge_kinds ok kinds in
-      let type_abbrevs = merge_type_abbrevs ota type_abbrevs in
-      let types = Flatten.merge_type_assignments ot types in
-      kinds, type_abbrevs, types
+    let kinds, type_abbrevs, types, functional_preds =
+      if hash = base_hash then
+        precomputed_kinds, precomputed_type_abbrevs, precomputed_types, functional_preds
```
gares, line 4032: The idea is that type checkig requires to merge some stuff. the precomputed_stuff is already merged.
I have some use cases in which the base used for type checking is not the base used for linking (a bit unsafe at the moment).
If the base is the same for both steps, the already merged stuff can be used. If not, one has to merge since the base changed and the precomputed values are wrong.

If it is also the case for functional_preds, follow the name schema/field positions a the other stuff.
If not, don't computed functional_preds in this if-then-else.

tests/sources/functionality/test14.elpi
```diff
@@ -0,0 +1,22 @@
...
+
+% Recall: q is not functional!
+% Note: in warren, the following rule is considered functional
+% since the input X should be ground, and therefore, q X does not
+% create any choice point for `r X Y` where Y can be assinged to 2 distinct
+% values.
+% In the implementation using the input mode of elpi this is no more true.
+f X Y :- q X, r X Y.
+
+main :- std.findall (f _ _) L, print L.
```
gares, line 22: assert the length

gares:
nice test suite. I think all the tests that pass you should run a findall and assert the length of the returned list is < 2

Issue #271, opened 2024-10-08
[spilling] weird spilling behaviour?


FissoreD:
is it wanted that:
`main :- q {p 1} X => q 1 3.`
is translated into:
`main :- (p 1 A1 , q A1 A0) => q 1 3.`
instead of
`main :- p 1 A1 ,  (q A1 A0 => q 1 3).`

FULL EXAMPLE:
```prolog
pred p i:int, o:int.
p 1 4.

pred q i:int, o:int.

main :- q {p 1} X => q 1 3. % shouldn't the goal fail?
```

gares:
spilling in negative position is broken, it is not even clear what one wants.
take
```
q A {p A 1} X.
```
should it be
```
q A Y X :- p A 1 Y.
```
?

In your example `p` has a constant input, so any position (but the current one) would do. But if its input was an argument of `q` then your grafting would be wrong.

FissoreD:
Mh, I see this is not clear.
However, it took me some time yesterday to see why my code behaved strangely :)
*Maybe we should prevent spilling when used in negative position?*

Moreover, when use spilling in positive position, but inside an anonymous lambda-abstraction, 
the error message is quite difficult to understand, at first sight...
```prolog
pred p o:int.
p 5.
main :- std.forall [1] (x\ print {p}).
```
*Shouldn't we force a sigma inside the anonymous function for the result of `p`, or again forbid the use of spilling in anonymous predicates.*

In fact, I think spilling is very good, but I complain that, when miss-used, one can spend a lot of time to understand the bug in the code if he is not familiar with spilling

gares:
hum, in that specific case it should make a sigma. Does it not? It may a bug because we don't know (yet) the types.
It surely does a sigma under a `pi x\`.

I fully agree it is currently a hack.

FissoreD:
> hum, in that specific case it should make a sigma. Does it not?

I post the code generated by the compiler
- `main :-  std.forall [1] (x\ print {p}).`  becomes $\to$ `main :- std.forall [1] (x\ (p (A0 x) , print (A0 x)).` 
 and after the first iteration it produces `Fatal error: Unification problem outside the pattern fragment.` since
 it tries to unify `p 5` with `p (A0 1)`.
 the correct compiled version shouldn't be `main :- std.forall [1] (x\ sigma A0\ (p A0 , print A0)).` ?
- In the case of:
[...]

PR #290, opened 2024-11-26
[determinacy] static analysis

FissoreD:
TODO:
- [x] func f i:... o:... (makes things more readable sometimes)
   NOTE: in the current parser, we already have a special case when a function has no output.
   In this case the arrow in the type signature is optional
- [x] Variadic in dtype (no special case for `,` )
    Ok, but in infer_fold for constants other than comma, we could stop if the call `is_wrong`, in comma
    we should continue since the cut could make the predicate again deterministic
- [x] review builtins -> todo: stdlib_map and stdlib_set: their main constructor carries a prediate which may impact the determinacy of calls to add, remove etc...
  A similar example below:
  ```prolog
  kind k type.
  type k0 (func -> fprop) -> k.
  func fst-proj k -> (func -> fprop).
  fst-proj (k0 F) F.
  pred g.

  func give-rel -> prop.
  give-rel g.

  func f.
  f :- fst-proj (k0 give-rel) X, X Y, Y. % Note that Y is not a function, but it is difficult to detect
  ```
- [x] cleanup env
- [x] What is the type of this p in this code:
   ```prolog
   pred p.
   func p.
   func g.
   g :- p.
   ```
   Should we consider that p is a function or a relation?
- [x] in pp_args in data.ml, should I check that all inputs are before all outputs?
- [x] change determinacy of Arithmetic tests
- [x] findall signature is not precise enough: it is functional and it returns a list of functional prop if the input is functional
- [x] Difficult case:
  ```prolog
  func map (func A -> B), list A -> list B.
  pred divisor i:int, o:int.

  func give-div list int -> (list(list int)).
  give-div L L1 :- map (x\y\x y) [map divisor L] L1. % here map is wrongly called in the list, 
                                                     % i.e. x should be rel
  ```
- [x] In the spilling (it is also true in master) the type inside Spill is wrong. E.g. if `t = (Spill({ty;it},_)` then `ty` is wrong, but `it.ty` is ok
- [x] In the follwing code, we should check the output of the anonymous functions to have the determinacy expected by its signature
   ```prolog
   func f.
  pred rel.

  f :-
    F = ((y\ y = rel) : (func -> fprop)),
    F R,
    R.
  ```
- ~[ ] For the overlapping check in Coq, be careful about the problematic-subterms as in [here](https://inria.hal.science/hal-04547069v2): the two non unifying terms `{{fun x => f (H x)}}` and `{{fun y => g (G y)}}` are both replaced with a uvar combined with a constraint relating the uvar with the original term. Wrt overlapping check, this will cause a compilation error of the DB...~
 Moved here: #328
- [x] error if 2 types only differ by the mode/functionality
- [x] check for grafting and cut (mut excl)
- [x] update Elpi.md
- [x] The indexing with depth 1 uses the MapOn index which is wrong when working with clause retrieval in the mut_excl...
- ~[ ] Not really linked with this PR: but DiscrTree updates the max_depth of terms to index, each time a new clause is indexed.
      The problem is that we can index a **local** clause with very deep arguments, when we quit the scope where this
      clause is loaded, the depths should be reset to the previous state...~ 
      Moved here: #329
- [x] Mut-excl: mut-excl should be corrected so that each indexed argument has a rigid occurrence of at least on freash pi-quantified variable. Example:
   ```prolog
   func f int -> int.
   func chain1 -> int.
   chain1 X :-
     pi x\ f (A x) 3 => f _ X.	
   func chain2 -> int.
   chain2 X :-
      pi x\ f (Z x) 2 => chain1 X.
   main :- std.findall (chain2 X) L, print L.
   ```
   There are two solution for `chain2 X` even though it is functional. The checker should prevent to load `f (Z x) 2` since its
   **indexed** input does not have a rigid occurence of bound variables
- ~[ ] The loading of hypothetical clauses that are variables may break functionality.~
   ```prolog
   pred p.
   func f o:func.
   pred q o:prop.
   q (f p).
   main :- q X, !, X => ... % Here the load of `X` breaks the func of `f` but we cant see it statically
   ```
    ~In the current implementation we warn the user. Maybe we can check (at least) mutual-exclusion.~
    ~[Here](https://github.com/FissoreD/elpi/blob/d6c6c83a76271c03e79e05197b8c939460cf1702/src/runtime/runtime.ml#L3923-L3926) we can check if the deterministic clauses returned by `clausify` are in `mut_excl` with the program~ 
    Moved here: #330


src/compiler/spilling.mli
```diff
@@ -0,0 +1 @@
+val main : Compiler_data.ScopedTerm.t -> Compiler_data.ScopedTerm.t
```
gares, line 1: missing header

gares:
please run `make menhir-strip-...`

src/compiler/determinacy_checker.ml
```diff
@@ -1,235 +1,762 @@
...
+      | Impl (true, _hd, t) -> infer ctx t (* TODO: hd is ignored *)
+      | Impl (false, _, t) -> infer ctx t (* TODO: this is ignored *)
+      | App (_, n, x, [ y ]) when F.equal F.eqf n || F.equal F.isf n || F.equal F.asf n ->
+          Format.eprintf "Calling inference for unification between \n - (@[%a@])\n - (%a)@." ScopedTerm.pretty x
+            ScopedTerm.pretty y;
+          let f1, f2 = (infer ctx x, infer ctx y) in
+          Format.eprintf "Inferred are \n - %a\n -%a@." pp_functionality f1 pp_functionality f2;
+          unify_func x y f1 f2;
+          Functional
+      | App (_, n, x, xs) when F.equal F.andf n ->
```
gares, line 510: It is probably better to add explicit And, Or... nodes in scoped term.
But in the meanwhile you have to check that the scope is Global.

gares:
overall looks good

src/builtin.ml
```diff
@@ -218,33 +220,37 @@ let core_builtins = let open BuiltIn in let open ContextualConversion in [
...
-  LPCode "type (=>) list prop -> prop -> prop.";
-  LPCode "type (==>) prop -> prop -> prop."; (* not really needed since the parser emits a => *)
-  LPCode "type (==>) list prop -> prop -> prop.";
+  LPCode "external symbol (:-)  : fprop -> fprop -> fprop.";
+  LPCode "external symbol (:-)  : fprop -> list prop -> fprop.";
+  LPCode "external symbol (,)   : variadic fprop fprop.";
+  LPCode "external symbol uvar  : A = \"core\".";
+  LPCode "external symbol (as)  : A -> A -> A = \"core\".";
+  LPCode "external symbol (=>)  : prop -> fprop -> fprop.";
+  LPCode "external symbol (=>)  : list prop -> fprop -> fprop."; (* HACS in TC to handle this*)
```
gares, line 229: @FissoreD WDYM?

src/runtime/runtime.ml
```diff
@@ -3187,12 +3187,15 @@ let rec claux1 loc get_mode vars depth hyps ts lts lcs t =
      claux1 loc get_mode vars depth hyps ts lts lcs
        (deref_appuv ~from ~to_:(depth+lts) args g)
   | Arg _ | AppArg _ ->
+      assert false |> ignore;
```
FissoreD, line 3190: Same here

src/runtime/runtime.ml
```diff
@@ -3187,12 +3187,15 @@ let rec claux1 loc get_mode vars depth hyps ts lts lcs t =
...
       error ?loc "The head of a clause cannot be flexible"
   | Builtin (c,_) -> raise @@ CannotDeclareClauseForBuiltin(loc,c)
   | (Lam _ | CData _ ) as x ->
      type_error ?loc ("Assuming a string or int or float or function:" ^ show_term x)
-  | UVar _ | AppUVar _ -> error ?loc "Flexible hypothetical clause"
-  | Nil | Cons _ -> error ?loc "ill-formed hypothetical clause"
+  | UVar _ | AppUVar _ -> 
+    assert false |> ignore;
```
FissoreD, line 3196: Same here

FissoreD (commented):
@gares

src/parser/grammar.mly
```diff
@@ -348,7 +361,7 @@ clause:
 
 attributes:
 | { [] }
-| EXTERNAL { [ External ] }
+| EXTERNAL; o = option(STRING) { [ External o ] }
```
FissoreD, line 364: Why the optional string? Isn’t the string already passed via the option(external_ref) above?
`| EXTERNAL; o = option(STRING) { assert (o = None); [ External o ] }` doesn’t break the code.
I suppose we can roll back this change.

src/compiler/compiler.ml
```diff
@@ -1518,340 +1474,328 @@ module Assemble : sig
...
+                in
+
+              let cl_overlap, index = R.Indexing.add1clause_overlap_runtime ~depth ~time:(runtime_tick ()) index p cl in
+              check_clause ~is_local:true ~depth ~loc:fresh_loc ~lcs:morelcs index cl cl_overlap p amap;
+              aux ~depth index l
+            with 
+            | CompileError _ as e -> raise e
+            (* TODO: this horrible catchall is for clause with flex head... *)
+            | Assert_failure _ -> aux ~depth index l end
+        | Builtin (And, l) ->
```
FissoreD, line 1759: Replace the Assert_failure with a more idiomatic error.
An error can be thrown by clausify1 if the clause has, for example, a flex head — which is not a fatal issue at compile time.

src/compiler/type_checker.ml
```diff
@@ -187,79 +347,81 @@ let silence_linear_warn f =
...
+    | Impl(b,t1,t2) -> check_impl ~positive ctx ~loc ~tyctx b t1 t2 ety
+    | Const(Global _,_,_ as c) -> check_global ctx ~loc ~tyctx c ety
+    | Const(Bound _,_,_ as c) -> check_local ctx ~loc ~tyctx c ety
     | CData c -> check_cdata ~loc ~tyctx kinds c ety
     | Spill(_,{contents = Phantom _}) -> assert false
-    | Spill(sp,info) -> check_spill ctx ~loc ~tyctx sp info ety
-    | App(Global _ as gid,c,x,xs) -> check_app ctx ~loc ~tyctx (c,gid) (global_type env ~loc c) (x::xs) ety 
-    | App(Bound lang as gid,c,x,xs) -> check_app ctx ~loc ~tyctx (c,gid) (local_type ctx ~loc (c,lang)) (x::xs) ety
-    | Lam(c,cty,t) -> check_lam ctx ~loc ~tyctx c cty t ety
+    | Spill(sp,info) -> check_spill ~positive ctx ~loc ~tyctx sp info ety
```
FissoreD, line 379: I'he double checked the role of the positive boolean in `check`.
It is still unused. I propose the following fix:
```suggestion
    | Spill(sp,info) -> 
      if not positive then error ~loc "Spilling in negative position is forbidden";
      check_spill ~positive ctx ~loc ~tyctx sp info ety
```
The typechecker raises an error if a spilling is found in negative position.
I think this partially fixes: [#271](https://github.com/LPCIC/elpi/issues/271)

src/parser/grammar.mly
gares: because I was not convinced of `symbol x : ty = "str"` that is `external "str" type x ty`.
I think we can remove it now, modulo updating the grammar in vscode and pygments.

PR #327, opened 2025-04-22
special constant for nil in dtree

FissoreD:


gares:
it seems no code generates it

FissoreD:
Forgot to push local commit. Now should be good

gares (approved):

gares:
CI needs update

FissoreD:
How can I use `trace-only` with two different flags: I want both `dev:disc-tree:depth-path` and `candidates`

Issue #331, opened 2025-04-29
Mutual exclusion: retroactive check between local clauses and catchall


gares:
We should track when a predicate has a catch-all rule and when it loads a rule that is mut_excl via a pi-quatified variable.

Currently, if there is a catch all rule followed by the other kind, we detect the problem, but not vice versa.

The simplest fix would be to track the pi-quantified one and error when a catch-all is accumulated (since the other case is already ok).

gares:
I think we should put in the "pred_info" metadata twp two booleans (maybe with a loc):
- has catch all
- has hyp clause on pi

and no matter in which order we update that info, we end up having both bools set to true, and in that case we error.

FissoreD:
Yep already done

FissoreD:
I'm testing and preparing a PR :)

FissoreD:
Also `uvar`

FissoreD:
```prolog
func copy int -> int.

copy _ 3 :- !.

main :-
    pi x\ copy x 4 => ....
```
There is bug in the code.
I expect a det-check error when loading `copy x 4` but there is not

PR #334, opened 2025-04-30
Performing a retroactive check on a rule that uses local clause loading and catchall

FissoreD:
Linked to #331

### DIFF
--- a/src/compiler/compiler.ml
+++ b/src/compiler/compiler.ml
@@ -1772,8 +1772,9 @@ end = struct
       | Const b -> min_depth <= b && b < depth
       | _ -> false
     in
-    let is_unif_var = function
+    let rec is_unif_var = function
       | AppUVar _ | UVar (_, _, _) | Discard | Arg (_, _)|AppArg (_, _) -> true
+      | App (h,x,xs) when h == Global_symbols.asc -> is_unif_var x && List.for_all is_unif_var xs
       | Nil|Const _|Lam _|App (_, _, _)|Cons (_, _)|Builtin (_, _)|CData _ -> false
     in
     (* Builds a query for the predicate p with args args *)
--- a/src/compiler/compiler.ml
+++ b/src/compiler/compiler.ml
@@ -1670,10 +1670,15 @@ end = struct
     let t  = todbl (depth,ctx) t in
     (!symb, !amap), t
 
-  let check_mut_excl state symbols pred_info cl cl_st p amap =
+  let check_mut_excl state symbols ~loc pred_info cl cl_st oc p amap : pred_info C.Map.t =
     let pp_global_predicate fmt p =
       let f = SymbolMap.global_name state symbols p in
       Format.fprintf fmt "predicate %a" F.pp f in
+
+    let preds_w_eigen_var_no_cut = ref C.Map.empty in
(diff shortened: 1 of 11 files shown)


src/compiler/compiler.ml
```diff
@@ -1743,32 +1767,48 @@ end = struct
       let b = aux t in
       (* Format.eprintf "Bool is %b@." b; *)
       b
+    in *)
+    let is_eigen_variable ~depth = function
+      | Const b -> 0 <= b && b < depth
```
gares, line 1772: This checks if the var is bound, not if it is the one bound by the pi.
```
main :-
  sigma x\
    foo x => foo _.
```
should be rejected

src/compiler/compiler.ml
```diff
@@ -1743,32 +1767,48 @@ end = struct
...
+      | _ -> false
+    in
+    let is_unif_var = function
+      | AppUVar _ | UVar (_, _, _) | Discard | Arg (_, _)|AppArg (_, _) -> true
+      | Nil|Const _|Lam _|App (_, _, _)|Cons (_, _)|Builtin (_, _)|CData _ -> false
```
gares, line 1777: `(uvar as X)` is not a catch all, but `(X as Y)` is (although not very useful).

src/compiler/compiler.ml
FissoreD: It is already the case that `foo x` would be rejected, in fact, whenever I encounter a sigma in the term, I subst each occurences bound by it with a fresh uvar, therefore the only positive const are names https://github.com/LPCIC/elpi/blob/a2354c61821849644ff6a1eb3ceb14dbba863add/src/compiler/compiler.ml#L1804-L1807

gares:
```
main :- pi x\ (bad :- (foo x) => foo x, bad) => bad.
```
Is this rejected? In the paper we do 1-pi, 1-=>, otherwise it is wrong.

FissoreD: You are right, this should be rejected and it is not.
In this commit 7e1b6638c5b152a3c0e5b4e0d71fb1eb2f1056ad, I adress the problem by changing in which interval the id of the eigenvariable should be (i.e. I add a min_depth).
I put two tests, one with your code ([test111.elpi](https://github.com/LPCIC/elpi/blob/7e1b6638c5b152a3c0e5b4e0d71fb1eb2f1056ad/tests/sources/functionality/test111.elpi)) and one which is a slightly different version ([test112.elpi](https://github.com/LPCIC/elpi/blob/7e1b6638c5b152a3c0e5b4e0d71fb1eb2f1056ad/tests/sources/functionality/test112.elpi)).

src/compiler/compiler.ml
FissoreD: Should be fixed here: 39dc4742a7321dfd82480f0389f409e9fc69c104. 
Tests [test113.elpi](https://github.com/LPCIC/elpi/blob/39dc4742a7321dfd82480f0389f409e9fc69c104/tests/sources/functionality/test113.elpi) and [test114.elpi](https://github.com/LPCIC/elpi/blob/39dc4742a7321dfd82480f0389f409e9fc69c104/tests/sources/functionality/test114.elpi)

gares (approved):

PR #339, opened 2025-05-12
fix cut is arg of a call + backtrace in error message

FissoreD:
The `cut` operator has special behavior when used as an argument of the `and` predicate.

For example, consider a list of comma-separated atoms like `a1, ..., an`. 
If `ai = cut`, then the execution of `a1, ..., ai` is functional, 
i.e., the choice points created by `a1, ..., a(i-1)` are discarded.

However, if `cut` is used as an argument of another predicate, 
we cannot be sure whether the choice points are discarded.

For instance, the atom `if C r !`, for a given condition `C` and relation `r`, 
should not be considered functional: the `!` does not behave the same 
way as it does in the `and` predicate.

---

We also add a backtrace to the message error to debug in an easier way the error
as an example run this `dune exec elpi -- -test tests/sources/functionality/test78.elpi`


gares:
In order to fix MC ci, you should replace the ssreflect package with 2 new ones, boot and order

gares (approved):

PR #348, opened 2025-07-27
Det check chr

gares:
constraints must be deterministic, since they can be resumed at any time
- we export the function
- we make declare_constraint require a func
- we make miscalled builtins an error
- we add syntax for variadinc func

todo:
- [x] test variadic func (not just builtin)
- [x] document variadic syntax
- [x] update error_messages.txt
- [x] fix test suite
- [x] CHR cannot return a !

### DIFF
--- a/tests/sources/variadic.elpi
+++ b/tests/sources/variadic.elpi
@@ -1,5 +1,5 @@
 func f int, int -> int.. .
-f N D R :- R is N div D.
+f N D R :- !, R is N div D.
 f N D R M :- f N D R, M is N mod D.
 
 main :- f 3 2 R, f 3 2 _ Q, R = 1, Q = 1.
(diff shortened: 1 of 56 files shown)


tests/sources/variadic.elpi
```diff
@@ -0,0 +1,5 @@
+func f int, int -> int.. .
+f N D R :- R is N div D.
+f N D R M :- f N D R, M is N mod D.
```
gares, line 3: this type checks fine, runs fine, but det check fails with an assertion

gares: it now works https://github.com/LPCIC/elpi/pull/348/commits/0215834c2550ef36433f492ecb5176b96aef80d0 but should somehow not require the cut

PR #349, opened 2025-07-28
DRAFT: elpi2html using react

FissoreD:
Works with: [https://github.com/FissoreD/elpi2html](https://github.com/FissoreD/elpi2html)
Online page: [https://fissored.github.io/elpi2html](https://fissored.github.io/elpi2html)

`program1` generates the JSON output.

To generate a JSON file, use the following command:

```bash
dune exec elpi -- a.elpi -print -test > FILE_NAME.json
```

Then, clean `FILE_NAME` by removing any output generated by the Elpi execution, keeping only the JSON part.

This process is currently messy and should be improved.

An example of a JSON file that can be opened in elpi2html is [this](https://github.com/FissoreD/elpi2html/blob/master/examples/x.json)

TODO:
- [x] builtins are printed as TODO

### DIFF
--- a/src/compiler/compiler.ml
+++ b/src/compiler/compiler.ml
@@ -2496,9 +2496,10 @@ let pp_program (pp : pp_ctx:pp_ctx -> depth:int -> _) fmt (compiler_state, { Ass
 ;;
 
 let pp_program_json (pp : pp_ctx:pp_ctx -> depth:int -> Fmt.formatter -> term -> unit) fmt (compiler_state, { Assembled.clauses=cl; signature; symbols }) =
+  let table = SymbolMap.compile symbols in
   let pp_ctx = {
     uv_names = ref (IntMap.empty, 0);
-    table = SymbolMap.compile symbols;
+    table;
   } in
 
   let cnt = ref 0 in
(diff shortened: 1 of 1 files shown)


gares:
I tried it and I see a problem here:

Screenshot from 2026-02-25 09-12-54

The source is 
```
compatible GX [X|XS] GY [Y|YS] [TX = TY | K] :-
 both_or_none (GX => term X TX) (GY => term Y TY),
 !,
 compatible GX XS GY YS K.
```
So I expect either to see `=>` or to see GX to be on top.

Also, all `:-` are red, while the old code was using that for neck cuts only.

src/compiler/compiler.mli
```diff
@@ -53,6 +53,7 @@ val optimize_query : query -> executable
 val relocate_closed_term : from:symbol_table -> to_:program -> term -> (term, string) Stdlib.Result.t
 
 val pp_program : (pp_ctx:pp_ctx -> depth:int -> Format.formatter -> term -> unit) -> Format.formatter -> program -> unit
+val pp_program1 : (pp_ctx:pp_ctx -> depth:int -> Format.formatter -> term -> unit) -> Format.formatter -> program -> unit
```
gares: Pp_program_js

PR #364, opened 2025-09-15
Add more information to `check_same_mode` error message

Janno:
This has already saved me a bunch of time in my quest to `func`-ify a code base with some duplicate declarations.

I tried to make the output nice but my knowledge of OCaml format strings is limited. I probably won't have time to tinker more with this so feel free to push changes to my branch before merging.


gares:
Thanks, the new message is indeed more informative

PR #403, opened 2026-04-11
API: accumulated files are visible in the API

### LINKED ISSUE #401: [wish] expose accumulate in parsed ast
`Parse.from_file` should return a list of `filename * Ast.t` and leave to the client to decide what to do when the same file is accumulated twice.

This may require to error if acumulate is done inside a name space, and in the middle of a file.

Maybe it is better to add a `:use foo.` that can only occur in the "header" of the file, solving all backward compatibility issues.

gares:
The APIs return all files recursively accumulated.

- [x] use parser output everywhere
- [x] RecoverStructure gives a list of programs
- [x] fix API: parse 1 file, get many units, deduplicate later if needed

Fix: #401

### DIFF
--- a/src/API.ml
+++ b/src/API.ml
@@ -65,7 +65,7 @@ let trace = set_trace
 
 let usage =
   Trace_ppx_runtime.Runtime.usage
-  type warning_id = Util.warning_id = LinearVariable | UndeclaredGlobal | FlexClause | ImplicationPrecedence
+  type warning_id = Util.warning_id = LinearVariable | UndeclaredGlobal | FlexClause | ImplicationPrecedence | NestedAccumulate
 
 let set_warn = Util.set_warn
 let set_error = Util.set_error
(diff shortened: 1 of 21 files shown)


gares:
@mysaajava This is the PR we talked about. Trocq fails this way:
```
2026-04-13T13:00:10.7704228Z [33m# [0mFile "./Param.v", line 46, characters 0-28:
2026-04-13T13:00:10.7704371Z [33m# [0mError: duplicate clause name log-level
```
Can you have a look?

gares:
All tests pass, but for trocq, so I think we should look into it since it is also the project that needed this feature the most.

gares:
I found the bug, it is on my side

MysaaJava:
Yes, it seemed that something was accumulated two times.
I can adapt Trocq to use a new accumulate interface if needed. You stuck with `accumulate` interface, right ?

gares:
ok, now trocq builds, but there are fixes in HB to be done.

src/API.ml
```diff
@@ -1340,12 +1349,17 @@ module Utils = struct
...
+    { ast =
+        [Program.Clause {
+          Clause.loc = loc;
+          attributes;
+          body = aux depth Util.IntMap.empty term;
+          needs_spilling = ();
+        }];
+      file_name = loc.Util.Loc.source_name;
+      deps = [];
+      digest = "";
```
gares: Fixme

gares:
@MysaaJava I've added a -deps command line flag to print deps.
The HB overlay worked, so I think I can merge this one.

I stayed with accumulate because the semantics it is still the one to concatenate file, i.e. if a accums b that accums c, a sees c. I may use another keyword when (if) a will only see b (and b will see c).

PR #405, opened 2026-05-04
Update Cygwin setup in GitHub Actions workflow

gares:
Replaced Cygwin setup command with opam init for package installation.


gares:
Thanks @kit-ty-kate your [suggestion](https://github.com/ocaml/opam/issues/6881) seems to have worked, even if the output was not really informative.
Maybe it should be properly documented.

PR #431, opened 2026-07-31
Fix =!=> for runtime hypotheses

VojtechStep:
When the LHS of =!=> was not comptime-known, the runtime implementation added a head cut instead of a tail cut.

Before the patch, running the new test with `dune exec elpi -- -test tests/sources/implbang_runtime.elpi -trace-on -trace-at run 1 4 -trace-only user:new-hyps` would give `user:new-hyps = (p A0) :- !, (A0 = 2).`, while the analogous `implbang.elpi` with comptime-known LHS gives the expected `user:new-hyps = (p A0) :- (A0 = 2), !.`

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,10 @@
 - Trace:
   - Add information about added clauses to Implication cards
 
+- Runtime:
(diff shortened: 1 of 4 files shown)


gares (approved):
Thanks

Issue #427, opened 2026-07-23
Interest in contributions to expand IO builtins?


shonfeder:
Hello! I love using elpi. Thanks for maintaining and continuing to evolve it!

I am working on a personal toy project and have occasion to considerably expand the builtins with bindings to ocaml functions form `Filename`, `Sys`, and `Unix` to expand its capabilities for OS interactions and file manipulations. The FFI extension mechanism is easy and works very well, so thanks for that!

To illustrate, at this point, I've just about a dozen declarations of this sort

```ocaml
    MLData (result (BuiltInData.poly "A") (BuiltInData.poly "B"));

    MLCode (Pred ("readdir",
      In  (BuiltInData.string, "DirPath",
      Out (result BuiltInData.(list string) BuiltInData.string, "Contents",
      Easy "get the contents of the directory at DirPath")),
      (fun path _ ~depth:_ ->
        try !: (Ok (Sys.readdir path |> Array.to_list))
        with Sys_error e -> !: (Error e))),
    DocAbove);
```

As I am accumulating this growing list of `MLCode` entries, it occurs to me it would be easy to contribute these back to the core elpi `builtin.ml`. Would you be interested in these contributions? (I can understand that you may well prefer not to bloat elpi with such features :) ).

gares:
yes, please submit this code!

on the style side, I did not use result but rather the diagnostic type (in many rocq elpi APIs).
Eg `coq.typecheck T  Ty ok` (success, assign Ty) and `coq.typecheck T  Ty (error Msg)` does not assign Ty. And `coq.typecheck T Ty D, if (D = error Msg) (print Msg) (print Ty)` is also a possibility. The former became quickly an idiom, and if the builtin is coded properly you don't even need to generate the error message if the user calls it with `ok` (and the call fails). There are also a few combinators, like `assert-ok!` and co, that work with this "skema".

I don't know how weird this looks to you. But having both result and diagnostic would certainly look weirder IMO.

Another (debatable maybe) advantage is that by hoisting the datum out of the `ok` node, one can easily return multiple values (without the pairing, tripling,... construction) and have them documented separately. eg `stat File Date Permissions More Diagnostic`.

shonfeder:
Great! Thanks for the guidance and confirmation. I'll hopefully have the Sys stuff ready for review finally tomorrow. I'm very happy to use the existing conventions, and wouldn't want to push for confusing duplication.

[...]

PR #434, opened 2026-08-01
Expand `Sys` builtins

shonfeder:
This aims at expanding the available builtins for interacting with the OS, as proposed in #427. I an opening this initial PR with just 6 additions, so I can gather feedback on whether the general approach and style is in line with your dev preferences.

I am happy to have this merged as is, if you are happy with it and would like contributions around this size (perhaps some tests should also be added? Could you point me to the right spot for those?). But I wold also be happy to gather feedback on the changes staged here, and then finish FFI for the following functions from `Sys`:

```
    external chdir : string -> unit = "caml_sys_chdir"
    external mkdir : string -> int -> unit = "caml_sys_mkdir"
    external rmdir : string -> unit = "caml_sys_rmdir"
    external getcwd : unit -> string = "caml_sys_getcwd"
```

### DIFF
--- a/src/builtin.elpi
+++ b/src/builtin.elpi
@@ -344,20 +344,23 @@ external func eof in_stream.
 % [gettimeofday T] sets T to the number of seconds elapsed since 1/1/1970
 external func gettimeofday -> float.
 
-% [file_exists Path] like [Sys.file_exists], is true if file at [Path]
-% exists
+% [file_exists Path] is like [Sys.file_exists]. It succeeds if the file at
+% [Path] exists
 external func file_exists string.
 
-% [is_directory Path Diagnostic] like [Sys.is_directory], is true if [Path]
-% is a directory
+% [is_directory Path Diagnostic] is like [Sys.is_directory]. It succeeds if
+% [Diagnostic = ok] and [Path] is a directory or [Diagnostic = error Msg]
+% and [Msg] describes an encountered system error.
 external func is_directory string -> diagnostic.
 
-% [remove Path Diagnostic] like [Sys.remove], is true unless [Diagnostic] is
-% [ok] and would otherwise assign to an [error Msg]
+% [remove Path Diagnostic] is like [Sys.remove]. It succeeds if [Diagnostic
+% = ok] and the file at [Path] is removed or [Diagnostic = error Msg] and
+% [Msg] describes an encountered system error.
 external func remove string -> diagnostic.
 
-% [rename OldPath NewPath Diagnostic] like [Sys.rename], is true unless
-% [Diagnostic] is [ok] and would otherwise assign to an [error Msg]
+% [rename OldPath NewPath Diagnostic] is like [Sys.rename]. It succeeds if
+% [Diagnostic = ok] and the file at [OldPath] is renamed to [NewPath] or
+% [Diagnostic = error Msg] and [Msg] describes an encountered system error.
 external func rename string, string -> diagnostic.
 
 % [getenv VarName Value] Like Sys.getenv_opt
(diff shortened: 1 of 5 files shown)


gares (approved):
It seems some api need recent ocaml. I think we could document that, and use ppx optcomp to have the only in newer versions

shonfeder:
Ah! I didn't see that about `Sys.is_regular_file`, we could skip adding that for now too? It's not really essential IMO. I'd be happy with just dropping that only sticking with stuff that is compatible with the current lower ocaml bound. WDYT?

gares:
Ok drop it

src/builtin.ml
```diff
@@ -129,16 +129,16 @@ let option_decl a = let open AlgebraicData in Decl {
   constructors = [
     K("none","",N,
       B None,
-      M (fun ~ok ~ko -> function None -> ok | _ -> ko ())); 
+      M (fun ~ok ~ko -> function None -> ok | _ -> ko ()));
     K("some","",A(a,N),
       B (fun x -> Some x),
-      M (fun ~ok ~ko -> function Some x -> ok x | _ -> ko ())); 
+      M (fun ~ok ~ko -> function Some x -> ok x | _ -> ko ()));
```
shonfeder, line 135: Looks like my editor trimmed trailing whitspace. Are you OK with me leaving that in the diff here, or would you like me to revert those changes?

gares (approved):

src/builtin.elpi
```diff
@@ -344,12 +344,48 @@ external func eof in_stream.
...
 
+% [file_exists Path] like [Sys.file_exists], is true if file at [Path]
+% exists
+external func file_exists string.
+
+% [is_directory Path Diagnostic] like [Sys.is_directory], is true if [Path]
+% is a directory
+external func is_directory string -> diagnostic.
+
+% [remove Path Diagnostic] like [Sys.remove], is true unless [Diagnostic] is
```
VojtechStep: I think "is true unless diagnostic is ok" seems to describe the opposite of its behavior, no?

VojtechStep: Hm, are you trying to say that the predicate fails if the user explicitly provides a `ok` and the internals try to unify it with a returned `error Msg`? That's a generic unification error and I think it's making the doc string more confusing to describe it like this.

I'd suggest rewording it to something like
```
[remove Path Diagnostic] like Sys.remove. On success sets Diagnostic to ok. On failure sets Diagnostic to (error Msg) with an error message Msg.
```
and the `ok`/`error` mismatch behavior follows from the usual handling of output arguments.

I'd also avoid describing queries as "true/false" in favor of the established "successful/failed".

shonfeder: Yes, the current wording is very awkward. Thank you for helping improve it!

I will swap out `true` for success/failure, thanks for alerting me to this convention. But I think your suggestion doesn't work as phrased, because there are two senses of success at work: if `successful` mean the predicate holds `failed` means that we backtrack, as in 

```prolog
% [assert! C M] takes the first success of C or fails with message M 
```

then, IIUC, `remove "foo/bar" D` will generally "succeed" regardless of the value assigned to `D`, and in fact `D` will only get a value of the call succeeds. So, `reomve "foo/bar" (error "file doesn't exist")` will "succeed" in this sense. But the value of `D` should match the underlying *system operation*, which is what your phrasing captures, but I think we are not using "success" in the same way it is elsewhere in the docs here.

How about this?

> [remove Path Diagnostic] like Sys.remove. [Diagnostic = ok] if the file at [Path] is removed, otherwise [Diagnostic = error Msg].

VojtechStep: Right, I'm conflating the two successes, it should have been "On successful removal, ..., else ...". Your rewording works for me too.

shonfeder: Nice, I'm reworking them to be like this currently.

```
% [remove Path Diagnostic] is like [Sys.remove]. It succeeds if [Diagnostic
% = ok] and the file at [Path] is removed or [Diagnostic = error Msg] and
% [Msg] describes an encountered system error.
external func remove string -> diagnostic.
```

I think it helps too to have the operation actually described, even if it seems obvious, since we can imagine a future where people unfamiliar withe the underlying ocaml are consulting the docs :D

shonfeder:
I am just reworking the doc comments. I have put together some crude tests in https://github.com/shonfeder/elpi/commit/f1e9d47b0c764ef60f27990a11b882ab4db063d0 . This is enough to convince me that all the additions are working as expected, but they are not suitable for inclusion in the repo as is, because they do no sandboxing of the file system mutations. I could not find an existing mechanism for this within your current test harness (tho running the tests thru dune rules rather than make could do this for us). If you're happy merging without the tests, that's OK with me. If you'd like the tests, please advise on whether you have a preferred sandboxing strategy.

shonfeder:
Should be ready for another look and merge if you're happy with it!

If tests are desired for this, dune cram tests could be a good fix, since all the functionality requires interaction with the OS environment.

gares (approved):

gares:
Now that I think about it twice, I think you should name all sys functions sys.*  A proxy can be provided if clients break due to the renaming of getenv, command, gettimeofday. Maybe the latter belongs to unix?

Maybe you can start putting the new code in sys and do the renaming of existing code in another pr.

gares:
Also, because of another merge, you should re-promote the tests.
It would also be nice to have a few lines in the CHANGES file.

shonfeder:
> Now that I think about it twice, I think you should name all sys functions sys.* A proxy can be provided if clients break due to the renaming of getenv, command, gettimeofday. Maybe the latter belongs to unix?
> 
> Maybe you can start putting the new code in sys and do the renaming of existing code in another pr.

I agree that this naming is better! That is actually what I am doing in the project I pulled these from, but I tried to adopt the existing convention here. I've made this change, leaving the existing predicates for a followup PR.

I've also regenerated the fixtures and rebased.

gares (approved):

PR #437, opened 2026-08-04
Don't depend on ANSITerminal on Windows, provide stub

VojtechStep:
I believe this should fix the Windows issues (~~but only CI will tell...~~ it said yes 🥹).

`ANSITerminal` becomes an optional dependency of the test suite, with a stub provided for cases where it's not available.


gares (approved):

gares:
Thanks you so much!

PR #438, opened 2026-08-04
Temporarily remove coq-trocq-std from CI

VojtechStep:
Trocq is apparently known to be currently broken [[1]]; among other things its opam file declares a dependency on
`"coq" {>= "9.0" & < "9.2"}`, while rocq-stdlib@master already dropped support for rocq < 9.2 [[2]].

Note also that ignoring the version range on "coq" doesn't help, since the "coq" package doesn't have a version for 9.2 in the opam repository (and doesn't exist in the rocq repository), as opposed to e.g. coq-core.

Removing Trocq makes the rest of the "test users" CI pass: https://github.com/VojtechStep/elpi/actions/runs/30910579757/job/91996039212

[1]: https://github.com/rocq-community/trocq/issues/86
[2]: https://github.com/rocq-prover/stdlib/pull/300


CohenCyril:
Trocq was not known to be broken, just the CI.

